PR #6, opened 2015-09-10
name_resolve should be an ArgCommand, not a Command

larsks:
Assuming that I am trying to use name_resolve correctly:

    >>> api.name_resolve('QmXfrS3pHerg44zzK6QKQj6JDk8H6cMtQS7pdXbohwNQfK')

It looks as if it needs to be specified as an `ArgCommand`, rather than
as a `Command`.


BrendanBenshoof:
+1 just spent half an hour figuring out we should make this change and you beat me to the commit.
I've tested this change and it works as intended.

This indicates we should also add a "did this command throw an exception when given sane input" baseline test.

amstocker:
Thanks!

PR #8, opened 2015-09-10
use defaults when host, port, or base == None

### LINKED ISSUE #7: ipfsApi.Client should use defaults when given None as a value
When `host`, `port`, or `base` is None, `ipfsAPi.client.Client` should use the default value rather than trying to work with None.  This makes the module easier to use; consider, for example, if I have a Python client that includes:

    import argparse

    def parse_args():
        p = argparse.ArgumentParser()
        p.add_argument('--host')
        p.add_argument('--port')
        p.add_argument('--base')
        return p.parse_args()

I would like to be able to do this...

    api = ipfsApi.Client(host=args.host, port=args.port, base=args.base)

...rather than requiring a complex set of conditionals to deal with various combinations of user-specified vs. user-wants-the-default values.

larsks:
This permits the caller to request the default value explicitly
by pasing None to either host, port, or base when instantiating
ipfsApi.client.Client.

Closes #7

### DIFF
--- a/ipfsApi/client.py
+++ b/ipfsApi/client.py
@@ -8,6 +8,9 @@ from .commands import Command, \
                       FileCommand
 from .exceptions import InvalidCommand
 
+default_host = 'localhost'
+default_port = 5001
+default_base = 'api/v0'
 
 
 class Client(object):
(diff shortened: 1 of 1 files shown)


amstocker:
Thanks for the pr, makes sense.

PR #15, opened 2015-09-12
raise ipfsApiError on http errors

### LINKED ISSUE #13: Commands should raise exceptions on error
If I call something like:

    value = api.cat('invalidhash')

I should not get back the string `{u'Message': u'invalid ipfs ref path', u'Code': 0}`.  I should get an exception.  Since we're using the `requests` library, a simple first step may be calling:

    response.raise_for_status()

...assuming that the HTTP API is returning useful status code.

larsks:
This causes `ipfsApi.http.HTTPClient` to raise an exception in response to
HTTP errors, rather than successfully returning an error message to the
caller.  This makes it easier to differentiate errors from successful
responses.

Closes #13.

### DIFF
--- a/ipfsApi/exceptions.py
+++ b/ipfsApi/exceptions.py
@@ -1,14 +1,17 @@
-class InvalidCommand(Exception):
+class ipfsApiError(Exception):
     pass
 
-class InvalidArguments(Exception):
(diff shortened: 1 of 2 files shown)


amstocker:
Thanks, this is something I definitely have thought about and this is a great solution.

PR #19, opened 2015-09-13
introduce unit tests to the python api

larsks:
this pr introduces a framework for unit testing as well as some basic unit tests.

104649d (Lars Kellogg-Stedman, 64 minutes ago)
   introduce unit tests

   this commit moves `test/tests.py` into `test/functional/tests.py` and
   introduces a new `test/unit/` directory containing unit tests.

   `tox.ini` is configured to only run unit tests (with test coverage
   reporting) by default; you can run the functional tests with:

       tox test/functional

   coverage reports go into `htmlcov-py27`, `htmlcov-py34`, etc.

   This also *stops* running pep8 tests by default, on the assumption that
   these make more sense when run explicitly in the pre-commit hook.

cfc12cc (Lars Kellogg-Stedman, 3 minutes ago)
   make functional tests pass

   the functional tests were failing because they were trying to compare the
   equality of lists, which were apparently not coming back in the right
   order.  I've solved this by simply sorting the returned results by name.

   there was also a python3 vs. python2 problem in `ipfsApi.filestream` that
   was tripping up these tests.


larsks:
I should also note that with this PR, all tests are currently passing, which means this would be a great time to hook things up to [travis-ci](https://travis-ci.org/) or something.

jbenet:
:+1: loving what i'm seeing happening all over this repo

amstocker:
I've actually never used travis-ci but I know everyone is using it, probably a good time to learn it

Issue #5, opened 2015-09-10
No pydoc documentation


larsks:
The methods on an API client object have no documentation:

    >>> import ipfsApi
    >>> api = ipfsApi.Client()
    >>> help(api.ls)

This yields documentation on the `partial`  class, which isn't terribly helpful.  It is also impossible to ask for documentation on these methods using `pydoc`:

    $ pydoc ipfsApi.Client.ls
    no Python documentation found for 'ipfsApi.Client.ls'

This all seems very developer-unfriendly.

BrendanBenshoof:
The way @amstocker  is building the commands is elegant, but make it hard to insert the pydoc documentation.

If we keep this style, we need to essentially add the documentation in the definitions at client.py and ensure it is carried through to the partial function (we can then over-write the pydoc for partial with our desired pydoc)

I'm not sure I like this solution.

BrendanBenshoof:
I'm starting a re-write of client.py so we can make both of these things work.

amstocker:
I like your idea for the rewrite of client.py and I think it makes a lot of sense.  However at this point we should get rid of the prepare method altogether and just do e.g. return self._add.request(client, path, **kwargs)

whereswaldon:
#46 added some pep257-style documentation, but I'll try to work pydoc in soon as well.

whereswaldon:
This issue essentially subsumes the #26 's "Write doc strings for all api methods"

PR #23, opened 2015-09-15
Make Static Methods

BrendanBenshoof:
This is a lot of changes all at once, and I really want a CR.

This ONLY converts all the commands to static methods. We still need to go back and document them all (I want help on that)

### DIFF
--- a/ipfsApi/client.py
+++ b/ipfsApi/client.py
@@ -45,150 +45,150 @@ class Client(object):
     def add(self, *args, **kwargs):
         """
         """
-        return FileCommand('/add')(self, **kwargs)
+        return FileCommand('/add')(self._client, *args, **kwargs)
 
     def cat(self, *args, **kwargs):
         """
         """
-        return ArgCommand('/cat')(self, **kwargs)
+        return ArgCommand('/cat')(self._client, *args, **kwargs)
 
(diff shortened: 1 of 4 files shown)


amstocker:
Thanks, this is great and will solve #5.  Im wondering though, should we just turn command, argcommand, and filecommand into functions at this point?

BrendanBenshoof:
Apathy is kicking in (that was a lot of difficult to automate typing :P ). As long as it works and adds minimal overhead, I am not too concerned by it.

Future modification and maintenance is unlikely to touch that bit of code very often.

amstocker:
Yeah it is a lot to type, I don't blame you haha.  Looks good to me.

amstocker:
Now we can hack away at copy http://ipfs.io/docs/commands/ into docstrings #5

jbenet:
I'm curious why this is changing? tbh, the previous structure seems more maintainable to me. can probably use a metaclass if you're really worried about duplicating ptrs (i prob wouldn't be)

jbenet:
(anyway i dont really care, whatever works for you guys :) :+1: )

amstocker:
The [issue](#5) was that calling ``help()`` on api methods didnt provide any helpful documentation.  There probably is a fancier way to do it but this is fine imo

PR #25, opened 2015-09-16
client rework

amstocker:
@BrendanBenshoof what do you think?


BrendanBenshoof:
I'm not a big fan of decorators (I think it can make it less clear what a function does), but it seems to solve the problem.

amstocker:
Yeah... sorry, sometimes I get carried away with python.  My intention is that the decorator just creates the class once, then the api methods are passed the request method of the Command.

PR #27, opened 2015-09-25
First pass at docstrings with doctest-ish examples

moreati:
I'm afraid there are still some functions/methods without any docstring I ran out of time to do.

The doctest looking examples are mixture of doctests that would pass if run as doctests (mainly pure functions that don't require setup/a connection) and interactive sessions I pasted/reindented.

It's still rough, but I figure it's better than nothing

### DIFF
--- a/ipfsApi/client.py
+++ b/ipfsApi/client.py
@@ -13,6 +13,9 @@ default_base = 'api/v0'
 
 
 class Client(object):
+    """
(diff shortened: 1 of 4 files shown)


amstocker:
This is awesome, thanks!  Is there a way to tell pytest not to check doctests in client.py?

moreati:
```
... #doctest: +SKIP
```
In each docstring, between the test call and the result should do it.

PR #29, opened 2015-10-05
Add support for specifying HTTPS or HTTP in the hostname.

ryepdx:
It's best to use SSL when pulling from a gateway, but the way this was written originally didn't allow for that. This is a backwards-compatible way to add support for that use case.


amstocker:
Cool!  Thanks.

BrendanBenshoof:
I've found the best solution to this problem is to force the user to provide the protocol header. That way we can also support more protocols in the future (ftp and sftp stand out as options)

Issue #28, opened 2015-09-28
get directory from hash


nmushegian:
Is there a simple inverse operation for `hash = ipfs.add(directory_path, recrusive=True)[-1]["Hash"]`?

amstocker:
Do you mean just getting an entire directory and saving it in to disk?  We currently don't support `ipfs get` but we definitely should soon (thanks for bringing this up!).  The result of `hash = ipfs.add(directory_path, recrusive=True)` is actually a list of all the individual files added to ipfs and their respective hashes, so the snippet you posted would result in an error.  In the mean time you will need to download and save each file by their individual hash.

nmushegian:
Sorry, I meant `hash = ipfs.add(directory_path, recrusive=True)[-1]["Hash"]`; the hash the of the root directory.

nmushegian:
Looking closer, it seems ipfs is similar to git in that directories are not objects. So there's no file layout metadata being added when I do `ipfs add -r ...` ?

amstocker:
`ipfs add -r` will build the directory structure such that each directory has a hash, so the overall structure is certainly maintained.  If you took the result of `hash = ipfs.add(directory_path, recrusive=True)[-1]["Hash"]` and called `ipfs.ls(hash)` it would give you a list of all the files and subdirectories you in the folder you added and their respective hashes.  However, note that things like unix privileges are not currently retained.  Hope that helps!

amstocker:
fixed by #30

PR #30, opened 2015-10-07
Filestream

### LINKED ISSUE #2: File Streaming
I haven't been able to get file streaming for a recursive add working yet (right now I do multiple requests and manually set up the links).  I'm pretty sure requests itself can't render multi-part/mixed bodies, which is how the IPFS http api expects it; https://github.com/ipfs/go-ipfs/blob/master/docs/implement-api-bindings.md#note-on-multipart--inspecting-requests, so we might have to use urllib3 or just do it manually.  If anyone knows how to do this, some help would be greatly appreciated.

amstocker:
This PR fixes file streaming so that file uploads and downloads are chunked.  Also added support for the get command to download files straight to disk.  Fixes #2 and #28.


amstocker:
Putting this up here for some CR.  I got frustrated trying to get the file streaming working with 2.6, but I will give it a try later this week.

BrendanBenshoof:
Why did you re-add all the dynamic function constructors? I thought we were getting away from those?

https://github.com/ipfs/python-ipfs-api/pull/30/files#diff-33e409056abcd2472116a6860bf25cafR43

amstocker:
I didn't re-add them, I just made the client class methods wrap calls to those functions so that we could add helpful docstrings.  Its basically the same thing the decorators did, but I agree with you that decorators are too confusing.

PR #31, opened 2015-10-19
support streaming for /cat option with stream=True

ikreymer:
Check for `stream=True` option for requests and just return the raw stream.

Useful for `cat`, especially when retrieving non-text data
eg:

```
r = api.cat('<hash>', stream=True)
data = r.read()
```

### DIFF
--- a/ipfsApi/http.py
+++ b/ipfsApi/http.py
@@ -71,6 +71,10 @@ class HTTPClient(object):
                                    params=params, files=files, **kwargs)
 
         if not decoder:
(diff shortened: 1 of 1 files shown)


amstocker:
This makes a lot of sense.  Thanks for the PR!

amstocker:
For this to work does `stream=True` have to be passed to requests?

ikreymer:
Yep, just checks to see if it was already sent to requests :)

PR #37, opened 2016-03-07
Fix for chained responses with newlines

micxjo:
This fixes the JSON decoder to handle chained responses which have newlines between the objects, such as for `api.ping(...)`.


amstocker:
looks good!  Thanks.

PR #38, opened 2016-03-16
Fix for get_pyobject() and add_pyobject() failing

ingokeck:
* Fixed encoding issue in parse_pyobject by decoding it as latin_1. This might relate to IPFS Bug #1817 https://github.com/ipfs/go-ipfs/issues/1817 
* Added  test for get_pyobject() and add_pyobject() in test/functional/tests.py which works fine.

NB: Other tests in tests.py fail, there might be an issue with the multipart code. At the moment test/functional/tests.py is not evaluated by tox because setup.cfg limits tests to tests/unit.


amstocker:
The encoding problem seems to be coming up a lot.  This looks great!  Thanks

Issue #34, opened 2016-01-08
_object_new undefined


jgraef:
I just tried calling `object_new`, which in turns tries to call `_object_new`. But `_object_new` is not defined. Patch attached.
[object_new.patch.txt](https://github.com/ipfs/python-ipfs-api/files/82771/object_new.patch.txt)

whereswaldon:
This should be resolved by #46

whereswaldon:
Since I believe that #46 fixed this, I'm closing it for now. Please post again if you experience more problems with it.

PR #46, opened 2016-07-22
Document, test, and improve functionality.

### LINKED ISSUE #45: not compatible with flattened multipart file uploads introduced in 0.4.0
Hi,

The 0.4.0 release of go-ipfs introduced a change to multipart file upload that breaks compatibility when adding directories (https://github.com/ipfs/go-ipfs/pull/2046)

Currently when adding a directory via py-ipfs-api  with `recursive=True`, the server adds a plain file with multipart encoding noise, e.g.: 

```
--7587cf2cd7c54e3393bc49f454e626ce
Content-Disposition: file; filename="a%2Fb"
Content-Type: multipart/mixed; boundary="debd9482563f4d0eb04c6dcfa5cef500"

--debd9482563f4d0eb04c6dcfa5cef500
Content-Disposition: file; filename="a%2Fb%2Fc"
Content-Type: application/octet-stream

abcd123

--debd9482563f4d0eb04c6dcfa5cef500--
--7587cf2cd7c54e3393bc49f454e626ce--
```

Instead of nested multipart items that are chunked by size, the API now expects one multipart boundary, with one part per file or directory.  Also, directories and symlinks are now expected to have the `Content-Type` of `application/x-directory` or `application/symlink`.

Here's an example of the expected input format: https://github.com/ipfs/go-ipfs/blob/8c4900b3b8341fdc4667e57d9e45fe73bad4d123/commands/files/file_test.go#L82

whereswaldon:
We've written unit and functional tests for most of the project. Additionally, we've implemented some chunks of the core spec that were missing. We have all of the Core API spec implemented with the exception of node_start, node_stop, repo_init, repo_config_get, and repo_config_set. We also have a working version of add, cat, and get. (Add does not currently handle filtering added files by a regular expression though.)

Resolves #45 and #34 .


whereswaldon:
Actually, we're not quite ready yet. Just realized this branch is missing something. Will open a new one when fixed.

whereswaldon:
Okay, fixed now.

lgierth:
This is great, thanks so much! :) I have only very little experience with python so I can't give much valuable feedback. @alexander255 @amstocker @jbenet what do you think?

parkan:
:+1: on this, we're currently shelling out to go-ipfs to do recursive adds

PR #51, opened 2016-08-10
Fix TravisCI and make compatible with go-ipfs v0.4.3rc2

### LINKED ISSUE #50: TravisCI is broken
I believe this is because it's testing Python 2.6, which #46 dropped support for. Additionally, the functional tests require a running IPFS daemon on the same machine. I believe we could make TravisCI spin one up with some finagling. I'll investigate that further.

whereswaldon:
TravisCI previously ignored the functional tests that required an IPFS daemon to be running. #46 enabled those tests, but did not configure TravisCI to run an IPFS daemon. Now, Travis will download the release version of go-ipfs v0.4.3rc2 and run it before running the tests, ensuring that the functional tests pass too.

There have been some changes to the HTTP API responses since go-ipfs v0.4.3rc1, so this PR also updates the tests to reflect those changes.

This will resolve #50

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,23 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
+before_script:
+  - eval "$(gimme 1.6)"
+  - mkdir $HOME/go
+  - export GOPATH="$HOME/go"
+  - export PATH="$PATH:$GOPATH/bin"
+  - echo $GOPATH
+  - echo $PATH
+  - go get -d github.com/ipfs/go-ipfs
+  - pushd . && cd $GOPATH/src/github.com/ipfs/go-ipfs && make install && popd
 install:
   - "pip install -r requirements.txt"
   - "pip install -r test-requirements.txt"
 script:
+  - ipfs init
+  - ipfs daemon &
   - tox -e pep8
   - tox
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,12 +1,24 @@
...
+  - python: "3.3"
+    env: TOXENV=py33
+  - python: "3.4"
+    env: TOXENV=py34
+  - python: "3.5"
+    env: TOXENV=py35
+before_script:
+  - wget "https://dist.ipfs.io/go-ipfs/v0.4.3-rc2/go-ipfs_v0.4.3-rc2_linux-amd64.tar.gz" -O /tmp/ipfs.tar.gz
+  - mkdir $HOME/bin
+  - pushd . && cd $HOME/bin && tar -xzvf /tmp/ipfs.tar.gz && popd
```
alexander255: I have never used Travis before: Are its scripts guaranteed to be run with `bash`?

alexander255:
Tested and verified that it indeed fixes all functional tests in `py-ipfs-api`. :+1:

.travis.yml
whereswaldon: I don't know, actually. I took the way that I did this from [the TravisCI Documentation](https://docs.travis-ci.com/user/installing-dependencies/#Installing-Projects-from-Source). Do you have a different way of doing this in mind?

alexander255: If it's in the official docs then it's fine. I just have seen too many thing fail because of bashisms to not ask. :wink:

alexander255:
Thanks for your contribution!

PR #52, opened 2016-08-10
Fix unit test `test.unit.test_utils.TestUtils.test_ls_dir` to not rely on the ordering of files on the file system

alexander255:
Apparently it was mostly luck that this has worked for some time.

### DIFF
--- a/test/unit/test_utils.py
+++ b/test/unit/test_utils.py
@@ -126,7 +126,14 @@ class TestUtils(unittest.TestCase):
         dirs = ['test2', 'test3']
         files = ['fsdfgh', 'popoiopiu']
         contents = (files, dirs)
-        self.assertEqual(utils.ls_dir(path), contents)
+        
+        # Sort items before comparing as the ordering of files returned by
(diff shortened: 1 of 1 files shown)


whereswaldon:
Changes make sense to me. How did you catch this? Do you think that there are other places in the tests where this could be happening?

whereswaldon:
+1 to this. I was just able to replicate this test failure. I can't believe this hadn't happened to me yet.

alexander255:
> How did you catch this?
Luck mostly.

> Do you think that there are other places in the tests where this could be happening?
Unless there other things similar to `os.listdir` that can cause this, I don't think so.

> I can't believe this hadn't happened to me yet.
Just a theory here: I believe that git will generally create pulled files in the order they were added to the index (which I believe is alphabetically), which will (generally) result in the inodes also being ordered that way on the disk.
If that is indeed the case then that would explain why this doesn't fail very often (particularly on always-clean installations such as Travis) and hasn't been noticed yet.

Anyway, thanks for verifying this!

PR #54, opened 2016-08-15
Update to build against go-ipfs v0.4.3-rc3

whereswaldon:
Since there's a new release candidate, the old build errors while trying to fetch release candidate 2.


whereswaldon:
Looks like this works, so I can't see a reason not to merge it.

PR #47, opened 2016-07-23
Add log and bitswap cmd

fredthomsen:
Added more API commands to improve this relative to js-ipfs-api as mentioned by #42.

Issue #42, opened 2016-06-04
Completeness?


jbenet:
What is the completeness of this api? in relation say to https://github.com/ipfs/js-ipfs-api ?

Many people are using ipfs from python and they've asked me this question lots. I think it would be good to have a spec completeness markdown checklist.

whereswaldon:
As far as I am aware, this is where things are. This is drawn from [here](https://github.com/ipfs/specs/blob/master/public-api/core/README.md).

There were a few items in the spec that I'm not sure how to implement (node_start and node_stop), and a few that people on IRC said to hold off on because they might have security implications.

### Core

- [x] version
- [ ] node
  - [x] id
  - [ ] start - not implemented due to unclear spec
  - [ ] stop - not implemented due to unclear spec
- [x] block
  - [x] get
  - [x] put
  - [x] stat
- [x] object
  - [x] data
  - [x] get
  - [x] links
  - [x] new
  - [x] patch
  - [x] put
  - [x] stat
- [x] refs
  - [x] local
- [ ] repo
  - [ ] init - not implemented due to security concerns
  - [x] stat
  - [x] gc
  - [ ] config get - not implemented due to security concerns
  - [ ] config put - not implemented due to security concerns
- [x] pin
  - [x] add
  - [x] ls
  - [x] rm

### Extensions

- [ ] name (ipns)
  - [ ] name publish
  - [ ] resolve
- [ ] dns
  - [ ] resolve
- [ ] tar
  - [ ] add
  - [ ] cat
- [ ] tour
  - [ ] list
  - [ ] next
  - [ ] restart
- [x] files
  - [x] add - see #49 
  - [x] cat
  - [x] get - see #48
- [ ] stat
  - [ ] bw
- [ ] mount
- [ ] bootstrap
  - [ ] add
  - [ ] list
  - [ ] rm
- [ ] bitswap - see #47 
  - [ ] stat
  - [ ] unwant
  - [ ] wantlist
- [x] log
  - [x] level
  - [x] tail

whereswaldon:
Actually, the above is a checklist from the wrong API? I'm confused about the discrepancies between [this](http://docs.ipfs.apiary.io/#reference/bitswap/get) and [this](https://github.com/ipfs/specs/blob/master/public-api/core/README.md).
@jbenet What is the difference in intention between those two specs?

### Here is the functionality of `py-ipfs-api` compared against the Apiary:

**NOTE**: Not all of the following are tested currently.
[...]

PR #55, opened 2016-08-15
Add bitswap commands

fredthomsen:
Adding bitswap command as suggested by #42.

Closing #47.

### DIFF
--- a/test/functional/tests.py
+++ b/test/functional/tests.py
@@ -492,7 +492,7 @@ class IpfsApiObjectTest(unittest.TestCase):
         self.assertEqual(result,
                 {'Hash': 'QmV4QR7MCBj5VTi6ddHmXPyjWGzbaKEtX2mx7axA5PA13G'})
 
-
+@skipIfOffline()
 class IpfsApiBitswapTest(unittest.TestCase):
 
     def setUp(self):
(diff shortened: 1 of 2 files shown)


whereswaldon:
The code looks good, thanks. Could you write a quick test for each of the bitswap functions? If you follow the general model of the tests in 'test/functional/tests.py', it should be pretty easy.

fredthomsen:
In-progress.  Should have mentioned that above.

fredthomsen:
Added the tests. Passing on my box but Travis seems to be taking a bit.

whereswaldon:
The entire IPFS organization on Github shares one Travis instance. We
queue up behind whatever builds are going on for each of the other
projects that IPFS hosts. It's unfortunate, and they're working to fix
it, but for now we have to wait... sigh.

fredthomsen:
Guess waiting until morning is key.  Life is good.

test/functional/tests.py
```diff
@@ -492,5 +492,28 @@ def test_object_patch_set_data(self):
         self.assertEqual(result,
                 {'Hash': 'QmV4QR7MCBj5VTi6ddHmXPyjWGzbaKEtX2mx7axA5PA13G'})
 
+
```
whereswaldon: @alexander255 just updated our testing framework to skip the functional tests if an IPFS daemon isn't running. It's a one-line change to make your tests have this feature. Take a look at [this class decorator](https://github.com/ipfs/py-ipfs-api/pull/57/files#diff-322d04fe8e037003724bccda9930e89eR49). If you could just throw that on this test class, I'm good to merge this.

alexander255:
:+1: on this change once that decorator has been added.

fredthomsen:
Done.

whereswaldon:
@fredthomsen Thanks a ton for putting this together! Sorry that we kept adding todos as you went.

PR #57, opened 2016-08-15
Make unit tests fail with a descriptive error message if local `ipfs daemon` node is not running

### LINKED ISSUE #53: Streamline functional/tests exec by doing a pre-check for a running ipfs daemon
If no ipfs daemon is present/running, the tox test runs/hangs for a lengthy period of time when entering the functional/tests portion.  It might be a more streamlined experience to perform a pre-check on the ipfs daemon state and fail fast rather than running/hanging.

alexander255:
Fixes #53

### DIFF
--- a/test/functional/tests.py
+++ b/test/functional/tests.py
@@ -1,14 +1,52 @@
 # _*_ coding: utf-8 -*-
 import os
-import shutil
 import json
+import shutil
+import socket
+import sys
 import unittest
+import logging
 
 import ipfsApi
 
 
+__is_available = NotImplemented
+def is_available():
+    """
+    Return whether the IPFS daemon is reachable or not
+    """
+    global __is_available
+    
+    if not isinstance(__is_available, bool):
+        s = socket.socket()
+        try:
+            s.connect((ipfsApi.default_host, ipfsApi.default_port))
+        except IOError:
+            __is_available = False
+        else:
+            __is_available = True
+        finally:
+            s.close()
+    
+    return __is_available
+
+
+def skipIfOffline():
+    if is_available():
+        return lambda func: func
+    else:
+        return unittest.skip("IPFS node is not available")
+
+
+def test_ipfs_node_available():
+    addr = "[{0}]:{1}".format(ipfsApi.default_host, ipfsApi.default_port)
+    assert is_available(), "Functional tests require an IPFS node to be available at: " + addr
+
+
+
 HERE = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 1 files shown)


test/functional/tests.py
```diff
@@ -1,14 +1,52 @@
...
+            __is_available = True
+        finally:
+            s.close()
+    
+    return __is_available
+
+
+def skipIfOffline():
+    if is_available():
+        return lambda func: func
```
whereswaldon: So I follow that this is a decorator, but not how it works when you're decorating the test classes. Could you explain the interaction to me? How does this work since `skipIfOffline` doesn't accept parameters?

whereswaldon:
I may not understand how it works, but it does work. I've tested it out on my machine. Thanks for putting this together!

test/functional/tests.py
alexander255: Decorators (in Python) are actually just pretty syntax around the magic `__call__(*args, **kwargs)` method.

Therefor:

```
@skipIfOffline()
def test_somehting():
    …
```

is the same as:

```
def test_something():
   …
test_something = skipIfOffline()(test_something)
```

were the `skipIfOffline()` part invokes the `skipIfOffline` function and (temporarily) stores the result. The result returned by `skipIfOffline` is in also a callable, that then gets invoked with the actual method as parameter. This second callable will then return *another* callable, that is later invoked as actual unit test by the test harness.

Definitely ask for clarification if this is hard to grasp for you (layers, layers, LAYERS). :wink:

alexander255:
Thanks for your feedback!

joshisa:
@alexander255 Out of curiosity, is the def is_available exposed as a part of the client api set.  If not, any thoughts on value in doing so?  I stumbled across an issue in another consuming project [oduwsdl/ipwb](https://github.com/oduwsdl/ipwb/issues/13) that showed interest in leveraging a daemon check api if it existed in py-ipfs-api .

alexander255:
@joshisa Thanks for pointing this out, I didn't think about the the broader issue at hand to be honest. The currently implementation is just a quick fix for the unit tests through, that just checks whether there is *something* running on the default IPFS host and port.
To formalize this we probably should add a new function (maybe `ipfsapi.connect()`?) that creates a new client instance and tries to call some API call (such as `version()`). On success you'd get the `ipfsapi.Client`-instance, on failure you'd get an exception. Do you think that'd be a proper solution?

whereswaldon:
That'd be alright I think. We could even check the version given against what we support and throw an error if it's unsupported.

PR #61, opened 2016-08-17
Fix swarm_connect typo

whereswaldon:
There was a simple typo in swarm_connect.


alexander255:
Nice catch!

Could we use this as an excuse to add unit tests for the swarm API? (i.e. Connect address, list addresses, disconnect address) It would be nice if we could prevent such a thing in the future.

whereswaldon:
Yeah, I suppose that we should. I'll get on that.

whereswaldon:
Closing because stale and, as far as I know, no longer relevant.

PR #62, opened 2016-08-17
Update get function to accept destination path

### LINKED ISSUE #48: Files get has no way to specify the name of the file once it is retrieved
In the IPFS CLI, you can use `ipfs get -o <hash>` to get a file and save it with a given name, but the `ipfsApi.Client().get()` function does not have this capability right now.

whereswaldon:
Now the `api.get()` function can accept a filepath that it will move the downloaded file to. This mimics the functionality of `ipfs get <hash> -o <path>`. This PR should resolve #48

### DIFF
--- a/ipfsApi/client.py
+++ b/ipfsApi/client.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import
 from . import http, multipart, utils
 from .commands import ArgCommand, Command, DownloadCommand, FileCommand
 from .exceptions import ipfsApiError
+import shutil
 
 default_host = 'localhost'
 default_port = 5001
@@ -201,16 +202,21 @@ class Client(object):
         return self._add.request(self._client, (), files,
                                  recursive=recursive, **kwargs)
 
-    def get(self, multihash, **kwargs):
+    def get(self, multihash, output_path='.', **kwargs):
         """Downloads a file, or directory of files from IPFS.
 
-        Files are placed in the current working directory.
+        Files are placed in the current working directory if
+        output_path is not specified. Otherwise, they are
+        moved to output_path.
 
         Keyword arguments:
         multihash -- unique checksum used to identify IPFS resources
+        output_path -- where to put the downloaded file
         kwargs -- additional named arguments
         """
-        return self._get.request(self._client, multihash, **kwargs)
+        self._get.request(self._client, multihash, **kwargs)
+        if output_path is not '.':
+            shutil.move(multihash, output_path)
 
     def cat(self, multihash, **kwargs):
         r"""Returns the contents of a file identified by hash, as a string.
--- a/test/functional/tests.py
+++ b/test/functional/tests.py
@@ -174,6 +174,21 @@ class IpfsApiTest(unittest.TestCase):
         os.remove(test_hash)
         self.assertNotIn(test_hash, os.listdir(os.getcwd()))
 
+    def test_get_file_with_rename(self):
+        self.api.add(self.fake_file)
+
+        test_hash = self.fake[0]['Hash']
+        local_file_name = "test_file.txt"
+
+        # Download the file and save it to the current working
+        # directory with the given name
+        self.api.get(test_hash, output_path=local_file_name)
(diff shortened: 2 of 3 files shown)


test/functional/tests.py
```diff
@@ -174,6 +174,21 @@ def test_get_file(self):
...
+
+        test_hash = self.fake[0]['Hash']
+        local_file_name = "test_file.txt"
+
+        # Download the file and save it to the current working
+        # directory with the given name
+        self.api.get(test_hash, output_path=local_file_name)
+        self.assertIn(local_file_name, os.listdir(os.getcwd()))
+
+        self.assertTrue(local_file_name in os.listdir(os.getcwd()))
```
alexander255: Why do the `assertIn` check effectively twice?

ipfsApi/client.py
```diff
@@ -201,16 +202,21 @@ def add(self, files, recursive=False, **kwargs):
...
 
         Keyword arguments:
         multihash -- unique checksum used to identify IPFS resources
+        output_path -- where to put the downloaded file
         kwargs -- additional named arguments
         """
-        return self._get.request(self._client, multihash, **kwargs)
+        self._get.request(self._client, multihash, **kwargs)
+        if output_path is not '.':
+            shutil.move(multihash, output_path)
```
alexander255: Is there really no better way to do this than to call `shutil.move`? What if the output file is very big and the output path is on a different file system? What if the current working directory isn't writable?

test/functional/tests.py
whereswaldon: Because I copied the test from the plain `test_get_file` and then didn't see the duplication of logic. Too fixated on checking that the file was in the directory to see that it was already checked for. Nice catch! I'll remove that.

ipfsApi/client.py
whereswaldon: Good question. I hadn't considered file system boundaries or writability, though I should mention that since the `get` function always downloads to the current working directory, I'd sincerely hope that it's writable. 
As I'm sure you can tell, I'm no Python pro. I know just enough that I was able to patch up the repo. I'm happy to help maintain this project, but I'm definitely still learning best-practices (and some features) in the language.
The other approach that I can see is to push this logic down into the `DownloadCommand` or maybe further into the `http` module. This could ensure that the file is never written to the CWD unless the user didn't specify another path. However, I'm not certain whether that logic is unique to the `get` function. I'll have to check on that.

Do you see any other alternatives?
Also, do you think that the module's current assumptions about the ability to write to the CWD represent a systemic design flaw?

Thanks so much for these code reviews, by the way.

alexander255: > The other approach that I can see is to push this logic down into the DownloadCommand or maybe further into the http module. This could ensure that the file is never written to the CWD unless the user didn't specify another path. …
> 
> Do you see any other alternatives?

That is exactly what I would have done, so no. Just pass the parameter down the stack. It tuns out the `http.py` modules [already has support for passing in a path](https://github.com/ipfs/py-ipfs-api/blob/master/ipfsApi/http.py#L153) (the `filepath` paramter). However it appears that the tar archive always contains a single file or directory with the IPFS hash as name that will then contain the actual contents, so you'll probably need to adapt the function somewhat to iterate over all members of the tar separately and extract them to the correct path.
Look at the implementation of `tarfile.TarFile.extractall` for this; specifically the part were it iterates over the members of the archive and modifies the file mode of directories: Something similar should be possible to do with the actual file names retrieved as part of the tar archive.

I hope this gets you started on tackling this. :smiley: 

> Also, do you think that the module's current assumptions about the ability to write to the CWD represent a systemic design flaw?

No, extracting to the CWD with the hash as file name is a sane default for the quick-'n'-dirty scripts that are likely to use our library. It just doesn't make a lot of sense in more "advanced" (as in: people you actually know what they do) scenarios.

> Thanks so much for these code reviews, by the way.

Your welcome! :smiley_cat:

whereswaldon:
The code has changed a lot since I started doing this, so I think it'd be better to start over from a clean slate. I'm going to close this PR and open a new one when I have something ready.

PR #64, opened 2016-08-27
Clean up the API client constructor and default params

fredthomsen:
Use appropriate default parameters here.


alexander255:
Looks good to me! :+1:

Issue #36, opened 2016-02-22
Doesn't handle UTF-8 text well


kseistrup:
The API seems to mangle UTF-8 encoded text in a way as if it had received ISO-8859-1 encoded bytes from IPFS and then transformed each byte to a multibyte UTF-8 sequence.

First we create a file with UTF-8 contents:
```sh
$ echo Frække frølår | ipfs add
added QmciJtwy4qy7nqzW4pDxhJbUhkpTjKiH5AAmkFLAWWfnmT QmciJtwy4qy7nqzW4pDxhJbUhkpTjKiH5AAmkFLAWWfnmT
$ ipfs cat QmciJtwy4qy7nqzW4pDxhJbUhkpTjKiH5AAmkFLAWWfnmT
Frække frølår
$
```
Next, let's try and fetch the same content from Python:
```py
>>> import ipfsApi
>>> api = ipfsApi.Client('127.0.0.1', 5001)
>>> res = api.cat('QmciJtwy4qy7nqzW4pDxhJbUhkpTjKiH5AAmkFLAWWfnmT').rstrip('\n')
>>> print(res)
FrÃ¦kke frÃ¸lÃ¥r
>>> print(res.encode('iso-8859-1').decode('utf-8'))
Frække frølår
>>>
```

candeira:
Reproduced under Python 3.5. The text comes out fine under Python2.7, so it's clear that the problem is a Python3 compatibility issue.

```
>>> res = api.cat('QmciJtwy4qy7nqzW4pDxhJbUhkpTjKiH5AAmkFLAWWfnmT').rstrip('\n')
>>> res
u'Fr\xe6kke fr\xf8l\xe5r'
>>> print res
Frække frølår
>>> res.encode('iso-8859-1').decode('utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xe6 in position 2: invalid continuation byte
>>> print res.encode('utf-8').decode('utf-8')
Frække frølår
```

kseistrup:
I'm sorry that I didn't mention the Python version earlier.

There's nothing wrong with the locale settings in my terminal:
```sh
$ locale | grep -v '"'
LANG=en_DK.UTF-8
LC_ALL=en_DK.UTF-8
$
[...]

PR #63, opened 2016-08-22
Generate Sphinx documentation with Napoleon

alexander255:
Quite a huge change, revised pretty much all API documentation we have had. Fixes #5 & #58.

Rendered result:
fs:/ipns/QmZ86ow1byeyhNRJEatWxGPJKcnQKG7s51MtbHdxxUddTH/Software/Python/ipfsapi/

PR #65, opened 2016-08-28
Revamp exception and unicode handling

alexander255:
Loosely depends on #63, since it also updates the documentation quite a bit.

### DIFF
--- a/ipfsApi/client.py
+++ b/ipfsApi/client.py
@@ -7,9 +7,8 @@ Classes:
 """
 from __future__ import absolute_import
 
-from . import http, multipart, utils
+from . import http, multipart, utils, exceptions
 from .commands import ArgCommand, Command, DownloadCommand, FileCommand
-from .exceptions import ipfsApiError
 
 DEFAULT_HOST = 'localhost'
 DEFAULT_PORT = 5001
@@ -165,7 +164,7 @@ class Client(object):
             >>> c.cat('QmTkzDwWqPbnAh5YiV5VwcTLnGdwSNsNTn2aDxdXBFca7D')
             Traceback (most recent call last):
               ...
-            ipfsApiError: this dag node is a directory
+            ipfsApi.exceptions.Error: this dag node is a directory
             >>> c.cat('QmeKozNssnkJ4NcyRidYgDY2jfRZqVEoRGfipkgath71bX')
             '<!DOCTYPE html>\n<html>\n\n<head>\n<title>ipfs example viewer</ …'
 
@@ -1356,7 +1355,7 @@ class Client(object):
             for r in res:
                 if "Extra" in r and len(r["Extra"]) > 0:
                     return r["Extra"]
-        raise ipfsApiError("empty response from DHT")
+        raise exceptions.Error("empty response from DHT")
 
     def dht_put(self, key, value, **kwargs):
         """Writes a key/value pair to the DHT.
--- a/ipfsApi/client.py
+++ b/ipfsApi/client.py
@@ -7,7 +7,7 @@ Classes:
 """
 from __future__ import absolute_import
 
-from . import http, multipart, utils, exceptions
+from . import http, multipart, utils, exceptions, encoding
 from .commands import ArgCommand, Command, DownloadCommand, FileCommand
 
 DEFAULT_HOST = 'localhost'
@@ -1078,7 +1078,7 @@ class Client(object):
                 '/ip4/101.201.40.124/tcp/40001/ipfs/QmZDYAhmMDtnoC6XZ … kPZc',
                 '/ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYER … uvuJ',
                 '/ip4/104.223.59.174/tcp/4001/ipfs/QmeWdgoZezpdHz1PX8 … 1jB6',
-                ...
+                …
                 '/ip6/fce3: … :f140/tcp/43901/ipfs/QmSoLnSGccFuZQJzRa … ca9z']}
 
         Returns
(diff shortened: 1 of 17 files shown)


alexander255:
This will also fix #36.

ipfsApi/client.py
```diff
@@ -1793,12 +1792,37 @@ def files_mv(self, source, dest, **kwargs):
     ###########
 
     @utils.return_field('Hash')
+    def add_bytes(self, data, **kwargs):
```
whereswaldon: Good idea. I hadn't thought to handle the different kinds of string in Python separately (b'' vs u''), but it makes sense.

ipfsApi/utils.py
```diff
@@ -3,143 +3,12 @@
...
 
-import io
-import json
 import mimetypes
 import os
 from functools import wraps
 
 import six
 
-from six.moves import cPickle as pickle
```
whereswaldon: I agree with getting rid of all of these wrapper functions; they were a little silly. :smile: 

On a more project-management note, I think that in the future we should try to do this kind of refactor in its own PR. Thoughts?

ipfsApi/encoding.py
```diff
@@ -42,48 +53,128 @@ class Json(Encoding):
...
+            result = json.dumps(obj)
+            if isinstance(result, six.text_type):
+                return result.encode("utf-8")
+            else:
+                return result
+        except (UnicodeEncodeError, TypeError) as error:
+            raise exceptions.EncodingError('json', error)
+
+
+class Pickle(Encoding):
```
whereswaldon: Why do we need to wrap this functionality?

ipfsApi/exceptions.py
```diff
@@ -1,32 +1,126 @@
-"""Defines the skeleton for exceptions.
 """
+The class hierachy for exceptions is::
 
+    Error
```
whereswaldon: :+1: 
I like this

ipfsApi/exceptions.py
```diff
@@ -1,32 +1,126 @@
...
+class Error(Exception):
+    """Base class for all exceptions in this module."""
     pass
 
 
-class InvalidArguments(ipfsApiError):
-    """Exception raised for invalid arguments."""
-    pass
+###############
+# commands.py #
```
whereswaldon: Is there really any need to mention source files in these comments? I'm not sure what we gain by doing so. Would you mind explaining?

ipfsApi/multipart.py
```diff
@@ -386,37 +386,43 @@ def _prepare(self):
...
-                names.append(('files', (short_name,
-                                        open(filepath, 'rb'),
-                                        'application/octet-stream')))
+                try:
+                    # add the file to those being sent
+                    names.append(('files', (short_name,
+                                            open(filepath, 'rb'),
+                                            'application/octet-stream')))
+                except OSError:
+                    # File might have disappeared between `os.walk()`
```
whereswaldon: Nice catch. Did this happen to you?

test/functional/tests.py
```diff
@@ -456,7 +456,7 @@ def setUp(self):
         self._olddir = os.getcwd()
         os.chdir(HERE)
         # Add a resource to get the stats for.
-        self.resource = self.api.add_str(u'Mary had a little lamb')
```
whereswaldon: Why remove the unicode marker? I feel like I'm missing something here.

tox.ini
```diff
@@ -20,5 +20,5 @@ commands =
   flake8 {posargs}
 
 [flake8]
-ignore = E222,E221,F403
+ignore = E222,E221,F403,E265
```
whereswaldon: Why do we now not require a space after pound symbols?

ipfsApi/utils.py
alexander255: Yes, it was mostly because I've been doing so much refactoring in the last couple days, I just batched the PRs by time of creation. Originally exception handling and unicode handling where even stuffed into one big commit, but I split it up later; probably at least should have done the same thing for the `pickle` en- and decoder. :expressionless: 

I'll try to better myself in the future. :blush:

ipfsApi/exceptions.py
alexander255: My thinking was to give a potential reader a general idea of "where is this used?", so that it's easier to figure out where this exception will probably originate from.

ipfsApi/multipart.py
alexander255: No, I just looked audited any code that can potentially cause an exception and the rest was experience and probably the fact that I read some stuff on Linux' `inotify` API some time ago.

test/functional/tests.py
alexander255: The `add_str` method still allows binary strings on Python 2 (but not on Python 3), because of backwards compatibility concerns and because it's, in my experience, very normal to use UTF-8 encoded binary strings in Python 2 instead of proper `unicode`.
By having the string be binary on Python 2 and text on Python 3, it's testing that the *de facto* preferred string type on each platform works properly. It's, in a way, the same reason why there used to be no `add_bytes` method at all.

tox.ini
alexander255: My main use case are "tagged comments" (an established expression may be available). In particular I've started marking python 2 compatibility code with `#PY2: …` comments, so that it's easier to find once we can finally drop that burden (probably in about 5 years :crying_cat_face:).

ipfsApi/encoding.py
alexander255: You mean the `Pickle` encoding class. Because it has such a nice symmetry with all the other encoding formats and because it allows one to use `decoder="pickle"` in the client API. :grin:

ipfsApi/exceptions.py
whereswaldon: Fair enough. Thanks

ipfsApi/encoding.py
whereswaldon: Yeah, I guess the API symmetry is probably better than how it would need to work if we didn't wrap it.

whereswaldon:
I am :+1: for merge

alexander255:
Thanks!

PR #66, opened 2016-08-30
Cleanup

alexander255:
Based on #65, because I'm too lazy to rebase directly on master and have all the documentation break and because some of the new exception stuff is used to make the final change (`connect()` function) actually useful.

Main feature of this branch is the removal of the `commands.py` layer of indirection. There were originally intended to be used directly, but, as noted in #5, this resulted in no documentation being attached to them. The (later added) workaround was to add another layer™, but even back then [it was suggested](https://github.com/ipfs/py-ipfs-api/issues/5#issuecomment-139311600) to remove the now redundant "commands". I happen to agree, since "commands" were the most confusing thing to understand for me when I first got to know the codebase.

Other changes include the removal of some Python 2.6 compatibility code in the unit tests and and API for checking whether the daemon is actually available and compatible when connecting (as suggested by @joshisa in #53).


whereswaldon:
Sorry that it's taking me so long to get to this. I'm planning on doing code review for this tomorrow.

joshisa:
@whereswaldon @alexander255 Will this lead to a new tagged release?   Pypi is pulling some pretty stale bits from March. ;-)

whereswaldon:
@joshisa  I think that we've done enough that it's time for a another tagged release. 

@alexander255 I've never published to PYPI before. Do we just need to tag the commit from Github, or is there some more complex process?

alexander255:
Since these commits do contain some breaking changes (I tried to keep it at a minimum), I think we should go ahead and get the renaming done once all the refactoring stuff is in. `ipfsApi`/`ipfs-api` would then continue to support IPFS 0.3.x on Python 2.x, while `ipfsapi` would support IPFS 0.4.x on Python 2.7+.

Publishing to PyPI is done by registering a project there (or have somebody give you access), then run `./setup.py upload`. Details here:
https://packaging.python.org/distributing/#uploading-your-project-to-pypi

Since we talked about renaming the package as well, we should register a new project for it and give all `py-ipfs-api` developers access to it (there is no formal organization support on PyPI). What do you think @whereswaldon?

whereswaldon:
@alexander255 I agree. The renamed package should have different support (>=2.7) and start at version 1.0.0 (or 0.1.0, depending on whether you think a client for an alpha technology should pretend to be stable).

test/functional/tests.py
```diff
@@ -46,6 +47,19 @@ def test_ipfs_node_available():
 
 HERE = os.path.dirname(os.path.abspath(__file__))
 
+class AssertVersionText(unittest.TestCase):
```
whereswaldon: Was this classname meant to be `AssertVersionTest`?

test/unit/test_multipart.py
```diff
@@ -95,18 +94,10 @@ def test_init_defaults(self):
...
-                          % (expected_type, generator.headers['Content-Type']))
-            if not re.match(expected_boundary_pattern, generator.boundary):
-                self.fail('Boundary does not match "%s"'
-                          % expected_boundary_pattern)
-        else:
-            self.assertRegexpMatches(generator.headers['Content-Type'],
-                                     expected_type)
-            self.assertRegexpMatches(generator.boundary,
-                                     expected_boundary_pattern)
+        self.assertRegexpMatches(generator.headers['Content-Type'],
```
whereswaldon: It feels so much better to lose the 2.6 special cases. These tests are all cleaner and better. It makes my soul happy.

whereswaldon:
Now that I've seen it without `commands.py`, I can tell that this is vastly superior. Why would we ever want to maintain that entire file when it added practically nothing? Thanks for this refactor, I think the code is in a far better place now. Other than one typo, I have no gripes with these changes. :+1: for merge whenever that typo is resolved (or whenever you tell me that it isn't a typo, whichever)

test/functional/tests.py
alexander255: I'm not sure anymore, but `AssertVersionTest` does sound better :grin:

alexander255:
Fixed the typo

alexander255:
Should I prepare another PR that does the renaming of the module?
Regarding version numbers: IMHO the first "new" version number should be `0.4.0` ("first release that targets IPFS v0.4"), after that we increase the revision number for each subsequent release. Breaking changes need to be postponed until the next IPFS daemon version (`0.5.x` in this case) is released.
What do you think?

whereswaldon:
@alexander255 I think that is the logical way to handler versioning. It will be less confusing for users if they don't have to keep up with a mapping of compatible version numbers. Another PR for renaming the module seems like the way to go.

PR #68, opened 2016-09-13
Test against go-ipfs v0.4.3-rc4

whereswaldon:


alexander255:
I tested it, CI tested it – works

alexander255:
Thanks for keeping track of this!

PR #70, opened 2016-09-15
Incrementally decode and return stream data sets

alexander255:
Includes some refactoring work to get rid of the "default encoding" thing in `HTTPClient`.
Also includes a minor breaking change in that `c.log_tail()` now returns a generator of dictionaries instead of a list of binary JSON strings.

### DIFF
--- a/ipfsapi/encoding.py
+++ b/ipfsapi/encoding.py
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
 """Defines encoding related classes.
 
 .. note::
@@ -7,6 +8,9 @@
 
 from __future__ import absolute_import
 
+import abc
+import codecs
+import io
 import json
 import pickle
 
@@ -18,9 +22,13 @@ from . import exceptions
 class Encoding(object):
     """Abstract base for a data parser/encoder interface.
     """
+    __metaclass__ = abc.ABCMeta
 
-    def parse(self, string):
-        """Parses string into corresponding encoding.
+
+    @abc.abstractmethod
+    def parse_partial(self, raw):
+        """Parses the given data and yields all complete data sets that can
+        be built from this.
 
         Raises
         ------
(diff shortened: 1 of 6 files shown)


whereswaldon:
I don't have the mental energy to look at this right now, but I'll try to in the next day or so. Just wanted to make sure that you knew I'd seen it.

ipfsapi/http.py
```diff
@@ -243,3 +241,7 @@ def session(self):
         yield
         self._session.close()
         self._session = None
+
+
+class HTTPClient(HTTPClientBase):
```
whereswaldon: Why declare this class? Just to have a concrete type inheriting from `HTTPClientBase`?

test/unit/test_encoding.py
```diff
@@ -35,16 +36,50 @@ def setUp(self):
...
+        data2_binary = six.b(json.dumps(data2))
+        assertPartialEqual(b"  " + data1_binary + b"  \r\n  " + data2_binary + b"  ", [data1, data2])
+        assertFinalizeEmpty()
+        
+        # String containing broken UTF-8
+        self.assertRaises(ipfsapi.exceptions.DecodingError,
+                          lambda: list(self.encoder_json.parse_partial(b'{"hello": "\xc3ber world!"}')))
+        assertFinalizeEmpty()
+    
+    def test_json_parse_incomplete(self):
```
whereswaldon: Thorough. :smile:

ipfsapi/encoding.py
```diff
@@ -18,21 +22,62 @@
 class Encoding(object):
     """Abstract base for a data parser/encoder interface.
     """
+    __metaclass__ = abc.ABCMeta
 
-    def parse(self, string):
-        """Parses string into corresponding encoding.
+    @abc.abstractmethod
+    def parse_partial(self, raw):
```
whereswaldon: I see now. You have to wait to throw the parsing exceptions. I'm basically leaving this comment for my own reference at this point.

ipfsapi/encoding.py
```diff
@@ -1,3 +1,4 @@
+# -*- encoding: utf-8 -*-
```
alexander255: It is only required for Python 2 and only if you want to use any non-ASCII characters in your script files (including comments, strings, …). Python 3 assumes utf-8 by default.

ipfsapi/encoding.py
alexander255: Yes, you call `parse_partial` as often as you want (collecting the results), then call `parse_finalize` to process and obtain any stuff that may be remaining the `encoding`'s buffers. For JSON that's just error handling (there is no reliable way to determine whether an error happened because of EOF or not).

ipfsapi/encoding.py
```diff
@@ -128,6 +242,20 @@ class Pickle(Encoding):
     """
     name = 'pickle'
 
+    def __init__(self):
```
alexander255: Yes, something along the lines of "actually we're lying and all you data will be buffered by this". :wink:

ipfsapi/http.py
alexander255: That shouldn't have sunk into this commit :blush: 
I was starting to work on `asyncio` support and apparently didn't stash these lines.

Your right in that it currently doesn't make sense to do this yet, will fix.

alexander255:
Done

Issue #49, opened 2016-08-03
Files add accepts unused filename filter


whereswaldon:
The recursive add functionality does not currently use the filter that it accepts to check the names of files. This should be implemented.

whereswaldon:
Compared to some other problems within the repo, this doesn't seem like a huge priority. I looked at what it might take to implement this, but I'm not going to try to do it just yet.

Implementation notes: This needs to be addressed on several levels:
- `ipfsApi.multipart.DirectoryStream._prepare` has access to the `fnpattern` because it's a property of the `DirectoryStream` object. However, it doesn't do anything with this.
- `ipfsApi.commands.FileCommand.request` does not accept an `fnpattern` explicitly, though `ipfsApi.commands.FileCommand.directory` does accept a `match` pattern that will be passed to the `DirectoryStream` (see above). Since `FileCommand.request` dispatches either `FileCommand.files` or `FileCommand.directory` based upon whether the add path is a directory, it needs to pass the pattern when it invokes `FileCommand.directory`, but not `FileCommand.files` (otherwise the request library gets angry about an unexpected keyword argument). 
- The `ipfsApi.client.add` function needs to accept a pattern (optionally) that is propagated down to the `DirectoryStream` if a directory is being added. Otherwise, it need not propagate it.

I'd welcome thoughts about a better approach to implementing this, as well as questions if my notes above don't make sense.

heaven00:
@whereswaldon close this ?

alexander255:
Yes, this has been fixed thanks to you :smiley:

PR #74, opened 2016-10-01
ENH: Files add accepts unused filename filter

heaven00:
#49

### DIFF
--- a/ipfsapi/client.py
+++ b/ipfsapi/client.py
@@ -104,7 +104,7 @@ class Client(object):
 
         self._client = self._clientfactory(host, port, base, **defaults)
 
-    def add(self, files, recursive=False, **kwargs):
+    def add(self, files, recursive=False, fnpattern='.*', **kwargs):
         """Add a file, or directory of files to IPFS.
 
         .. code-block:: python
@@ -121,13 +121,15 @@ class Client(object):
             A filepath to either a file or directory
         recursive : bool
             Controls if files in subdirectories are added or not
+        fnpattern : str
+            *fnmatch* pattern of filenames to keep
 
         Returns
         -------
             dict: File name and hash of the added file node
         """
         body, headers = multipart.stream_filesystem_node(
-            files, recursive, self.chunk_size
+            files, recursive, fnpattern, self.chunk_size
         )
         return self._client.request('/add', decoder='json',
                                     data=body, headers=headers, **kwargs)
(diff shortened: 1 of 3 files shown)


heaven00:
@whereswaldon I am not sure about the regex `.*` as default, it can be changed to `[a-zA-Z-0-9.-_]` that also works

```
>>> import re
>>> re.match('[a-zA-Z-0-9.-]*', 'ipfs.py')
<_sre.SRE_Match object; span=(0, 7), match='ipfs.py'>
>>> re.match('[a-zA-Z-0-9.-]*', 'go-ipfs.py')
<_sre.SRE_Match object; span=(0, 10), match='go-ipfs.py'>
>>> re.match('[a-zA-Z-0-9.-_]*', 'go-ipfs.py')
<_sre.SRE_Match object; span=(0, 10), match='go-ipfs.py'>
>>> re.match('[a-zA-Z-0-9.-_]*', 'go_ipfs.py')
<_sre.SRE_Match object; span=(0, 10), match='go_ipfs.py'>
>>> re.match('[a-zA-Z-0-9.-_]*', 'go_ipfs.py')
```

alexander255:
Thanks @heaven00 for looking into this! This thing is really overdue…

After reading through the diff, I can see one (and only one) problem: You seem to have mixed up regular expressions (in particular PCRE) and `fnmatch`. These aren't the same thing; in particular: `fnmatch:* == pcre:.*`, `fnmatch:*ipfs* == pcre:.*ipfs.*` and `fnmatch:.ipfs. == pcre:\.ipfs\.`. Also check out the [Python documentation](https://docs.python.org/3/library/fnmatch.html) on this.

Finally regarding you comment about `[a-zA-Z-0-9.-_]*`: Yes, it will match `ipfs.py`, but what about `Things to do: A short list.txt` or `That über cool 寿司 recipe.doc`? `*` would match both, while the example you provided (there is no equivalent of that in fnmatch actually) is both more complex and would not work.

Note to self: We should probably extend fnmatch functionality in the future (not this PR), by:
 * Allowing multiple patterns to be passed as a list
 * Allowing a raw regular expression to be passed instead

heaven00:
@alexander255 yes I have, thanks for the info :)

will change `re.match` to `fnmatch.fnmatch` that will also change the patterns back to `*`

P.S. thanks for the quick review, will push out the changes asap

alexander255:
That was very soon indeed! Code now looks good to me! :+1:

heaven00:
YaY! :D

PR #77, opened 2016-10-22
Support glob patterns instead of simple fnmatch patterns for `add`

alexander255:
Today I, for the first time, really tried to use the new (not yet shipped) `fnpattern` parameter and was pretty disappointed about its real-world utility. After trying different things I settled on the attached code for what I believe to be a much more useful interface:
The `pattern` parameters now expects a glob instead of a mere fnmatch pattern and can to things such as `**/meta/*/zone/**/*.jpg`. Also it is possible to pass more than one matching pattern by passing a list of strings instead of a single string parameter. Finally, because it happened to naturally emerge from the implementation, one can also pass fully fledged regular expressions (or anything else implementing `.match`) along with the _glob_ patterns in that list simply by pre-compiling them.


alexander255:
Oh and I got rid of the top-level `fake_dir` and tweaked the one test that used it to use `test/functional/fake_dir` instead.

whereswaldon:
I apparently missed the notification email for this. I'm so sorry that it took so long for me to notice. Next time this happens, just mention me in a comment and I'll get right on it. I'm looking at the changes today.

ipfsapi/multipart.py
```diff
@@ -328,20 +377,30 @@ class DirectoryStream(BufferedGenerator):
...
                  chunk_size=default_chunk_size):
         BufferedGenerator.__init__(self, directory, chunk_size=chunk_size)
 
-        self.directory = directory
+        self.patterns = []
+        patterns = [patterns] if isinstance(patterns, str) else patterns
+        for pattern in patterns:
+            if isinstance(pattern, str):
+                self.patterns.append(glob_compile(pattern))
+            else:
```
alexander255: Allowing *anything that implements the `.match` "interface"* here is indeed considered "Pythonic", it also allows programmers to implement custom matching function that do not use `re` at all. (Note however that the term Pythonic isn't very well defined and highly subjective; other people may interpret it very differently.)
Also, from a practical point-of-view, it very hard to reliably detect whether something actually is a compiled regular expression object or not, since `type(re.compile(""))` returns `'_sre.SRE_Pattern'`, which is (1) private (starts with an underscore) and (2) virtual (cannot be accessed directly from Python code: There is no `_sre.SRE_Pattern` class accessible from Python).

ipfsapi/multipart.py
```diff
@@ -317,6 +316,56 @@ def body(self):
             yield chunk
 
 
+def glob_compile(pat):
+    """Translate a shell glob PATTERN to a regular expression.
+
+    This is almost entire based on `fnmatch.translate` source-code from the
```
alexander255: Thanks!

ipfsapi/multipart.py
```diff
@@ -317,6 +316,56 @@ def body(self):
...
+                res = res + '.*'
+            else:
+                res = res + '[^/]*'
+        elif c == '?':
+            res = res + '[^/]'
+        elif c == '[':
+            j = i
+            if j < n and pat[j] == '!':
+                j = j + 1
+            if j < n and pat[j] == ']':
```
alexander255: `ipfsapi.multipart.glob_compile("[]")` returns `re.compile('^\\[\\]\\Z(?ms)$', re.MULTILINE|re.DOTALL)` for me. Are you sure about this result?

alexander255:
I just wanted to start complaining about nobody looking at this, so it's pretty much perfect timing I'd say. :grin: 
Stuff like this happens way too often to me as well. Next time I'll try to be more noticeable.

ipfsapi/multipart.py
```diff
@@ -317,6 +316,56 @@ def body(self):
...
+        i = i + 1
+        if c == '/' and len(pat) > (i + 2) and pat[i:(i + 3)] == '**/':
+            # Special-case for "any number of sub-directories" operator since
+            # may also expand to no entries:
+            #  Otherwise `a/**/b` would expand to `a[/].*[/]b` which wouldn't
+            #  match the immediate sub-directories of `a`, like `a/b`.
+            i = i + 3
+            res = res + '[/]([^/]*[/])*'
+        elif c == '*':
+            if len(pat) > i and pat[i] == '*':
```
metagriffin: i love what you guys are doing, and would relicense to LGPL -- can you open a ticket?

ipfsapi/multipart.py
whereswaldon: Okay, thanks for the explanation!

ipfsapi/multipart.py
alexander255: @metagriffin: I was hoping I could provoke this response. Thanks already!

whereswaldon:
I think this would pass Travis is you rebased on top of master.

PR #78, opened 2016-12-06
Fix tox formatting error

whereswaldon:
CI is erroring out because of this single-line format error.
@alexander255 Go ahead and merge this unless you have a reason not to.


alexander255:
Thanks for looking into this stuff!

PR #79, opened 2016-12-06
Switch to show the build status of master

whereswaldon:
Previously, the badge was reporting the outcome of the last build, no matter what it was. If the last PR was good, it would say the build was passing, even if master wasn't.


alexander255:
Sounds reasonable, I'll merge

PR #80, opened 2016-12-07
Add project that uses py-ipfs-api

machawk1:


alexander255:
Thanks for submitting this! Looks like a really useful project!

whereswaldon:
This is sweet! Thanks for adding it.

PR #86, opened 2017-05-27
added key_gen function to client

sunnya97:
Useful for creating new keys in order to push multiple objects to the IPNS namespace using name_publish.

### DIFF
--- a/ipfsapi/client.py
+++ b/ipfsapi/client.py
@@ -743,22 +743,43 @@ class Client(object):
         args = (name,)
         return self._client.request('/resolve', args, decoder='json', **kwargs)
 
-    def key_gen(self, keyName, type='rsa', size='2048', **kwargs):
+    def key_list(self, **kwargs):
+        """Returns a list of generated public keys that can be used with name_publish
+
+        .. code-block:: python
+
+            >>> c.key_list()
+            [{'Name': 'self',
+              'Id': 'QmQf22bZar3WKmojipms22PkXH1MZGmvsqzQtuSvQE3uhm'},
+             {'Name': 'example_key_name',
+              'Id': 'QmQLaT5ZrCfSkXTH6rUKtVidcxj8jrW3X2h75Lug1AV7g8'}
+            ]
+             
+        Returns
+        -------
+            list : List of dictionaries with Names and Ids of public keys.
+        """
+        return self._client.request('/key/list', decoder='json')['Keys']
+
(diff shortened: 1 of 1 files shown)


ipfsapi/client.py
```diff
@@ -743,6 +743,42 @@ def resolve(self, name, recursive=False, **kwargs):
         args = (name,)
         return self._client.request('/resolve', args, decoder='json', **kwargs)
 
+    def key_gen(self, keyName, type='rsa', size='2048', **kwargs):
```
alexander255: Please make `type` a mandatory argument (we don't want to hard-code a crypto type that may eventually become obsolete) and `size` an *int*.

ipfsapi/client.py
```diff
@@ -743,6 +743,42 @@ def resolve(self, name, recursive=False, **kwargs):
...
+            >>> c.key_gen('exampleKeyName')
+            {'Name': 'exampleKeyName',
+             'Id': 'QmQLaT5ZrCfSkXTH6rUKtVidcxj8jrW3X2h75Lug1AV7g8'}
+
+        Parameters
+        ----------
+        keyName : str
+            Name of the new Key to be generated. Used to reference the Keys.
+        type : str
+            Type of key to generate
```
alexander255: Please list the currently possible key types (rsa, ed25519) here so that they appear in the API documentation

ipfsapi/client.py
alexander255: Also please write `key_name` instead of `keyName` – we use snake_case everywhere else, no point in introducing any inconsistency here.

ipfsapi/client.py
```diff
@@ -743,6 +743,42 @@ def resolve(self, name, recursive=False, **kwargs):
...
+        size : str
+            Bitsize of key to generate 
+
+        Returns
+        -------
+            dict : Key name and Key Id
+        """
+
+        existingKeys = self._client.request('/key/list', decoder='json')['Keys']
```
alexander255: See the above comment about snake_case.

alexander255:
Yes, @sunnya97, please rebase onto master… most of the CI errors should now be fixed.

Also please run `tox -e pep8` fix the errors reported by the style-checker (I know that thing is quite annoying, sorry 'bout that) before pushing again, so that the build won't fail because of that.

Thanks so much for helping us keeping the library updated! :grinning:

sunnya97:
Updated!

PR #88, opened 2017-05-28
Fix for one of the tests

TheKysek:
This PR fixes "test_repo_gc"


alexander255:
Thanks for finding this! I've added your commit to #87 (which contains a bunch of other testing fixes as well).
You'll still receive attribution for this fix :slightly_smiling_face:

alexander255:
Also feel free to open a new PR if you have any other fixes (CI-related or not) that you think could be useful!

PR #90, opened 2017-06-01
Release version 0.4.1

alexander255:
It's been a while since our last release and a few things have accumulated. We probably should do a release at this point, no?


lgierth:
I generally love releases :):)

alexander255:
Done publishing! See #91 for a quick run-down of how that looks like. :smiling_imp:

PR #93, opened 2017-06-21
Add recursive and nocache options to name_resolve API call.

dmzimmerman:
The current Python API call for IPNS name resolution (name_resolve) does not support the recursive and nocache options that are supported on the command line; this very minimal pull request adds them, with their defaults set to false (as on the command line) so that existing Python code that relies upon the current API will work without modification.


alexander255:
Thank you for this useful little addition! Looks good!

PR #100, opened 2017-07-24
IPFS v0.4.10

alexander255:
Need to check if shutdown test correctly interactions with the CI

### DIFF
--- a/ipfsapi/client.py
+++ b/ipfsapi/client.py
@@ -759,7 +759,7 @@ class Client(object):
         -------
             list : List of dictionaries with Names and Ids of public keys.
         """
-        return self._client.request('/key/list', decoder='json')['Keys']
+        return self._client.request('/key/list', decoder='json', **kwargs)
 
     def key_gen(self, key_name, type, size=2048, **kwargs):
         """Adds a new public key that can be used for name_publish.
@@ -786,12 +786,6 @@ class Client(object):
         -------
             dict : Key name and Key Id
         """
-
-        existing_keys = self.key_list()
-
-        for key in existing_keys:
-            if key['Name'] == key_name:
-                return key
 
         opts = {"type": type, "size": size}
         kwargs.setdefault("opts", opts)
--- a/ipfsapi/client.py
+++ b/ipfsapi/client.py
@@ -794,6 +794,55 @@ class Client(object):
         return self._client.request('/key/gen', args,
                                     decoder='json', **kwargs)
 
+    def key_rm(self, key_name, *key_names, **kwargs):
+        """Remove a keypair
+
+        .. code-block:: python
+
+            >>> c.key_rm("bla")
+            {"Keys": [
+                {"Name": "bla",
+                 "Id": "QmfJpR6paB6h891y7SYXGe6gapyNgepBeAYMbyejWA4FWA"}
+            ]}
+
+        Parameters
+        ----------
+        key_name : str
+            Name of the key(s) to remove.
+
+        Returns
+        -------
+            dict : List of key names and IDs that have been removed
+        """
+        args = (key_name,) + key_names
+        return self._client.request('/key/rm', args, decoder='json', **kwargs)
+
+    def key_rename(self, key_name, new_key_name, **kwargs):
+        """Rename a keypair
+
+        .. code-block:: python
+
+            >>> c.key_rename("bla", "personal")
+            {"Was": "bla",
+             "Now": "personal",
+             "Id": "QmeyrRNxXaasZaoDXcCZgryoBCga9shaHQ4suHAYXbNZF3",
+             "Overwrite": False}
+
+        Parameters
+        ----------
+        key_name : str
+            Current name of the key to rename
+        new_key_name : str
+            New name of the key
+
+        Returns
+        -------
+            dict : List of key names and IDs that have been removed
+        """
+        args = (key_name, new_key_name)
+        return self._client.request('/key/rename', args, decoder='json',
+                                    **kwargs)
+
     def name_publish(self, ipfs_path, resolve=True, lifetime="24h", ttl=None,
                      key=None, **kwargs):
         """Publishes an object to IPNS.
--- a/ipfsapi/client.py
+++ b/ipfsapi/client.py
@@ -1013,6 +1013,83 @@ class Client(object):
 
         return self._client.request('/pin/ls', decoder='json', **kwargs)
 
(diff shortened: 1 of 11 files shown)


alexander255:
I think I've added support for all non-experimental items from
https://ipfs.io/blog/27-go-ipfs-0-4-10/
`+` a short-but-useful script for running tests in a CI-like environment locally.

Need review of the code now, then we can do a release 🙂

ipfsapi/client.py
```diff
@@ -1013,6 +1058,83 @@ def pin_ls(self, type="all", **kwargs):
...
+            Path to the new object to be pinned
+        unpin : bool
+            Should the pin of the old object be removed? (Default: ``True``)
+
+        Returns
+        -------
+            dict : List of IPFS objects affected by the pinning operation
+        """
+        #PY2: No support for kw-only parameters after glob parameters
+        if "unpin" in kwargs:
```
whereswaldon: This is a nifty little trick. Thanks!

test.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Stop daemon if it is still running
+kill -9 ${DAEMON_PID} 2>/dev/null
+if [ $? -eq 0 ];
+then
+	echo "IPFS daemon was still running after test!" >&2
+fi
+
+# Clean up IPFS directory
+rm -rf "${IPFS_PATH}"
```
whereswaldon: This is a handy script! That will help enormously with testing while it's in development.

test/functional/tests.py
```diff
@@ -648,5 +740,44 @@ def test_bitswap_unwant(self):
         result = self.api.bitswap_unwant(key='QmZTR5bcpQD7cFgTorqxZDYaew1Wqgfbd2ud9QqGPAkK2V')
         self.assertTrue(result is not None)
 
+# Run test for `.shutdown()` only as the last test in CI environments – it would be to annoying
+# during normal testing
```
whereswaldon: :+1:

test/functional/tests.py
```diff
@@ -648,5 +740,44 @@ def test_bitswap_unwant(self):
...
+@skipUnlessCI()
+@pytest.mark.last
+@pytest.mark.xfail(reason="Daemon does not shut down on Travis for some reason")
+class IpfsApiShutdownTest(unittest.TestCase):
+    def setUp(self):
+        self.api = ipfsapi.Client()
+    
+    def _is_ipfs_running(self):
+        # Test only running on Linux CI – use procfs directly
+        for pid in filter(lambda n: n.isdigit(), os.listdir("/proc")):
```
whereswaldon: I think this won't work on non-unix OSes. Do we care?

whereswaldon: @alexander255 ^^

whereswaldon:
@alexander255 I have two questions:
1. Is there a reason that our shiny new CI-like testing script can't be the one that we use for Travis? We could replace the current 4 lines of `script` in the `.travis.yml` with that, which would ensure that we'd notice if the script ever got out of date.
2. There's a test that almost definitely cannot work on Windows (because of a chronic lack of `/proc`). Do we consider testing on Windows a big enough priority to try to fix that? I don't own a Windows computer, so I'm not currently in a good position to experiment with other solutions. Maybe we should just open an issue for this and merge as-is?

test.sh
mguentner: Could you please add something like

```
[[ -d ${IPFS_PATH} ]] && rm -rf "${IPFS_PATH}"
```

Also ``rm -rmi`` instead of ``rm -rf`` gives you a chance to abort a deletion of your precious $HOME

alexander255: @mguentner: The `${IPFS_PATH}` is hard-coded to something like `$(pwd)/test-tmp` at the top of the script. No user IPFS data will be touched when running the daemon as part of this setup, so your user `$IPFS_PATH`/`~/.ipfs` will never be harmed by this in any way.

test/functional/tests.py
alexander255: Actually it works on Linux and on Linux only. When I initally wrote it that was OK, because I only intended for it to be run on Travis (where it still doesn't work for unknown reasons…).
Then I added the `./test.sh` script for testing this test and later decided to add it to a commit – which means the test isn't Travis-exclusive anymore.

I think I'll rewrite this to be cross-platform (including Windows). Not doing so will just result in technical depth that needs to be cleaned up down the road. 😉

alexander255:
> Is there a reason that our shiny new CI-like testing script can't be the one that we use for Travis?

No. Done!

> There's a test that almost definitely cannot work on Windows

Fixed but untested as I do not have an appropriate test environment. If we want to take Windows support seriously, we should should probably look into Windows CI testing environments. There's probably other bugs there as well…

Thanks for all the review sofar! :+1:

test/functional/tests.py
```diff
@@ -648,5 +739,86 @@ def test_bitswap_unwant(self):
...
+    def setUp(self):
+        self.api = ipfsapi.Client()
+        self.pid = int(os.environ["PY_IPFSAPI_TEST_DAEMON_PID"])
+    
+    @staticmethod
+    def _pid_exists(pid):
+        """
+        Check whether pid exists in the current process table
+
+        Source: https://stackoverflow.com/a/23409343/277882
```
whereswaldon: I shouldn't be surprised by how gross this solution is, but somehow I still am. Thanks for finding a way!

whereswaldon:
I don't know how feasible it is for us to maintain compatibility with an operating system that neither of us use. Maybe if a Windows-using maintainer comes on board at some point? In any event, all of my concerns have been addressed.

test/functional/tests.py
alexander255: 😁

alexander255:
Problem with that is that, at least in the no-so-enterprisey area we're dealing with here, developers don't seem to use Windows themselves very often.
Just like Python 2, Windows is somethings that you have to *support*, not something you *use*. :wink:

Anyway I looked for Windows CI testing and the list of hosted solutions is quite managable:

 * AppVeyor
 * Buildkite
 * GitLab CI

Any preferences?

whereswaldon:
Will we be able to use GitLab CI without migrating the project hosting?
AppVeyor and Buildkite both offer free accounts to OSS, but to get one on Buildkite you have to have an email back-and-forth with the support team. That makes me somewhat inclined to use AppVeyor. I have no experience with either, so if you've had a particularly good or bad experience with these, I'm happy to defer to your judgment.

PR #102, opened 2017-08-02
Add support and tests for several `.add()` options

### LINKED ISSUE #84: add only_hash=True
I am trying to get the hash of an existing file without adding to the IPFS.
ipfs add myfile --only_hash  seems to work
but ipfs_api.add(myfile, only_hash=True):
TypeError: request() got an unexpected keyword argument 'only_hash'

Am I doing something wrong?
Thanks.
Pat.

alexander255:
This also refactors `IpfsApiTest` to use py.test `assert` and to properly clean
up pins after each test.

Fixes #84.

### DIFF
--- a/test/functional/tests.py
+++ b/test/functional/tests.py
@@ -98,6 +98,10 @@ class IpfsApiTest(unittest.TestCase):
     fake_file  = 'fake_dir/fsdfgh'
     fake_file_only_res = {'Name': 'fsdfgh',
                           'Hash': 'QmQcCtMgLVwvMQGu6mvsRYLjwqrZJcYtH4mboM9urWW9vX'}
+    fake_file_dir_res = [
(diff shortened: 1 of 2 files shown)


test/functional/tests.py
```diff
@@ -162,10 +166,21 @@ def setUp(self):
 
         # Makes all of the diff visible if the hashes change for some reason
         self.maxDiff = None
+        
+        self.pinned = set(self.api.pin_ls(type="recursive")["Keys"])
 
     def tearDown(self):
         os.chdir(self._olddir)
 
+    def _clean_up_pins(self):
```
whereswaldon: Good idea here! Dunno why we didn't do it sooner.

alexander255:
Thanks! Let's do a release! 🙂

PR #105, opened 2017-08-07
Release version 0.4.2

alexander255:
“Restore feature-parity with go-IPFS 0.4.10”

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -1,7 +1,7 @@
 Since releasing new versions is currently a somewhat complicated task, the current procedure
-(02.06.2017) will be outlined in this document.
+(07.08.2017) will be outlined in this document.
 
-All of this has only ever been tested on Debian 9 (Linux).
+All of this has only been tested on Debian 9 & Fedora 26 (Linux).
 
 # Prerequirements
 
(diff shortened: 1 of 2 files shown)


alexander255:
This is now uploaded. The version is `0.4.2.post1` because you cannot currently fix the description on PyPI without uploading a new version: https://github.com/pypa/warehouse/issues/2170

mguentner:
Could you please tag `0.4.2.post1` / push the tag to github?

alexander255:
There is no code change between `0.4.2` and "`0.4.2.post1`" was never released as such. The only difference were adding a few characters to the `README.md` to get `pandoc` to convert it to syntactically valid restructedText. 

Is there any reason you cannot use the `0.4.2` release tag from GIT?

mguentner:
Yes. The reason is that the tag has not been pushed:

https://github.com/ipfs/py-ipfs-api/tags

alexander255:
That's weird… Did GIT stop uploading tags automatically in some recent change!?

Anyway, it should be fixed now. Tag `0.4.2` is now a thing.

mguentner:
Yeah, thank you!

PR #106, opened 2017-08-08
Improve testing configuration and switch to `flit`

alexander255:
Highlights:

 * Replace the `test.sh` shell script by a `test/run-tests.py` python script that is called from `tox` (not the other way around anymore)
    - The previous configuration only “happened to work” on CI because of how we parallize builds
 * Drop all the `.coveragerc-*` files
 * Parallize code style checks and integration tests on CI
 * Avoid downloading dependencies more than once on CI

… and **switch to `flit` for packaging**. (I'm so sold on this tool, waaaaay better than setuptools! 😄)

(All of this is post-0.4.2 ofc.)


README.md
```diff
@@ -159,9 +157,9 @@ You can submit bug reports using the [GitHub issue tracker](https://github.com/i
 ### Pull requests
 
 Pull requests are welcome.  Before submitting a new pull request, please
-make sure that your code passes both the [pep8](https://www.python.org/dev/peps/pep-0008/) formatting check:
+make sure that your code passes both the [code formatting](https://www.python.org/dev/peps/pep-0008/) formatting check:
```
whereswaldon: Maybe we should only say `formatting` once?

flit.ini
```diff
@@ -0,0 +1,38 @@
...
+
+author       = py-ipfs-api team
+author-email = 
+home-page    = https://github.com/ipfs/py-ipfs-api
+keywords     = ipfs storage distribution development
+license      = MIT License
+
+description-file = README.md
+
+# Unfortunately these currently need to be duplicated from `requirements.txt`
```
whereswaldon: I wish we could only include these in one place and simply parse them out of that one place in every other context, but I don't see how to do it with .ini files in the mix. Sigh...

test/run-tests.py
```diff
@@ -0,0 +1,124 @@
...
+
+# Spawn IPFS daemon in data directory
+DAEMON = subprocess.Popen(["ipfs", "daemon"])
+os.environ["PY_IPFSAPI_TEST_DAEMON_PID"] = str(DAEMON.pid)
+
+# Collect the exit code of `DAEMON` when `SIGCHLD` is received
+# (otherwise the shutdown test fails to recognize that the daemon process is dead)
+if os.name == "posix":
+	import signal
+	signal.signal(signal.SIGCHLD, lambda *a: DAEMON.poll())
```
whereswaldon: Didn't realize you could make variadic closures like that, though I suppose it makes sense. Nice trick!

whereswaldon:
On a random note, at time of writing the Travis build status (on the GitHub side) was still pending, but the Travis build UI shows a passing build. It's weird.

README.md
alexander255: Yeah, I just wanted to drop the term `PEP-8` because the `pep8` tool now has renamed itself to `pycodecheck` because Guido basically told them they were not faithful enough to the standard, so it may be misleading that they name themselves after it. (I have started to really dislike the tool personally, but that's another story…)

flit.ini
alexander255: Filed issue about this: https://github.com/takluyver/flit/issues/128

alexander255:
RE build status: I was watching it for about half an hour yesterday (before going to bed) and it was stuck in "waiting for queue" for some reason.

PR #115, opened 2018-01-11
Fix pre-commit script

suzaku:
There's no `pep8` environment configured in `tox.ini`.
We should use `codestyle` instead.

PR #116, opened 2018-01-11
Fix travis build

suzaku:
In my other PR #115, Travis complains "File exists" when executing `mkdir $HOME/bin`.

I think the installation of `IPFS` should be configured in some global *before run* hook, but I haven't found such configuration in the Travis documentation yet.

So here's a quick fix.


whereswaldon:
Yeah, this will work for the time being. Thank you!

Issue #114, opened 2017-12-05
Expose pub sub API


zbyte64:
Pub sub functionality is now available through the IPFS HTTP API, would like the client to support such functionality.

whereswaldon:
You're right that we should have that! We'll take a stab at it soon unless you're interested in adding it?

suzaku:
@whereswaldon Can I help with this? I'm just starting to learn more about IPFS, but I've written a lot of Python code, so I guess I can do this if you give me some documentation about this API.

whereswaldon:
@suzaku We'd love the help. The HTTP API is documented here: https://ipfs.io/docs/api/#apiv0pubsubls
You may have to scroll down to reach it. Find the heading `api/v0/pubsub/ls`, which is the first HTTP endpoint related to Pub/Sub.
It ought to be possible to use a lot of the existing HTTP-request-related code to do this. Look at how existing functions are implemented as a guide.
Thanks!

suzaku:
@whereswaldon 

Thanks, I'll take a look at the documentation.
By the way, just out of curiosity, why isn't `requests` used directly?

whereswaldon:
There were certain types of request that required more complex management
than others. We actually used to do a lot more in the http logic. We
switched to requests in the last year, I think. Before that, I managed a
bunch of generators that wrote multipart http request bodies. It was a
mess.
Anyway, I think we still need the abstraction that the HTTPClient class
provides. I could be wrong though. If you still think we could use requests
directly after making this change, we can discuss a refactor.

suzaku:
@whereswaldon

Thanks for replying.
The additional methods for `pubsub` API are easy to add, but I can't run the tests on my laptop.
There seems to be some kind of compatibility issue with `ipfs`, for example, I would see the following error:

```python
___________________ IpfsApiTest.test_add_filepattern_from_dirname ____________________

self = <tests.IpfsApiTest testMethod=test_add_filepattern_from_dirname>

    def test_add_filepattern_from_dirname(self):
        res = self.api.add(self.fake_dir, pattern=self.pattern)

        try:
>           assert self._sort_by_key(self.fake_dir_fnpattern_res) == self._sort_by_key(res)
E           AssertionError: assert [{'Hash': 'Qm...test2/fssdf'}] == [{'Hash': 'Qmb...'Size': '22'}]
[...]

PR #123, opened 2018-04-27
Pubsub

slandau3:
Hello everyone. Apologies if there are mistakes in this PR. First time contributing to a project.

This is a pull request for issue #114, the pubsub api.

I have implemented python functions for all pubsub methods: ls, pub, sub, and peers. I have also included a few test cases (that all pass) for the aforementioned methods. Note that the test case for "peers" does not actually test anything for I can't predict what peers the function will return.

Another note for the sub api. I didn't want to introduce another dependency for a webhook (?) so I simply returned a stream for that function. In the documentation for the pubsub_sub function I made it clear if a user wants to obtain published content they simply need to iterate over the generator (of the stream) that's returned. 

ex:
```python

sub = pubsub_sub('testing')

# publish something to the testing topic
pubsub_pub('testing', 'message')

# will block until stream is closed so this should be run on a separate thread
for word in sub:
    print(word)   # will print anything published to the 'testing' topic. In this case word will be data about 'message'

```

Once again this is my first time contributing to an open source project so I apologize for any mistakes. I ran `tox` for both codestyle and test cases.

### DIFF
--- a/test/functional/tests.py
+++ b/test/functional/tests.py
@@ -862,26 +862,23 @@ class IpfsApiPubSubTest(unittest.TestCase):
 
 
         # get the subscription stream
-        sub = self.api.pubsub_sub(topic)
+        with self.api.pubsub_sub(topic) as sub:
 
-        # make sure something was actually returned from the subscription
-        assert sub is not None
+            # make sure something was actually returned from the subscription
+            assert sub is not None
 
-        # publish a message to topic
-        self.api.pubsub_pub(topic, message)
+            # publish a message to topic
+            self.api.pubsub_pub(topic, message)
 
-        # get the message
-        sub_data = next(sub)
+            # get the message
+            sub_data = sub.read_message()
 
-        # end the subscription
-        sub.close()
+            # assert that the returned dict has the following keys
+            assert 'data' in sub_data
+            assert 'topicIDs' in sub_data
 
-        # assert that the returned dict has the following keys
-        assert 'data' in sub_data
-        assert 'topicIDs' in sub_data
-
-        assert sub_data['data'] == expected_data
-        assert sub_data['topicIDs'] == expected_topicIDs
+            assert sub_data['data'] == expected_data
+            assert sub_data['topicIDs'] == expected_topicIDs
 
     def test_pubsub_ls(self):
         """
@@ -892,15 +889,12 @@ class IpfsApiPubSubTest(unittest.TestCase):
         expected_return = { 'Strings': [topic] }
 
         # subscribe to the topic testing
-        sub = self.api.pubsub_sub(topic)   
+        with self.api.pubsub_sub(topic) as sub:
 
-        # grab the channels we're subscribed to
-        channels = self.api.pubsub_ls()
+            # grab the channels we're subscribed to
+            channels = self.api.pubsub_ls()
 
-        # unsubscribe (cleanup)
-        sub.close()
-
-        assert channels == expected_return
+            assert channels == expected_return
 
     def test_pubsub_peers(self):
         """
--- a/test/functional/tests.py
+++ b/test/functional/tests.py
@@ -851,6 +851,10 @@ class IpfsApiPubSubTest(unittest.TestCase):
         # get the 
         sub_data = next(sub)
 
+        # assert that the returned dict has the following keys
+        assert 'data' in sub_data
+        assert 'topicIDs' in sub_data
+
         assert sub_data['data'] == expected_data
         assert sub_data['topicIDs'] == expected_topicIDs
 
--- a/test/functional/tests.py
+++ b/test/functional/tests.py
@@ -889,12 +889,16 @@ class IpfsApiPubSubTest(unittest.TestCase):
         expected_return = { 'Strings': [topic] }
 
         # subscribe to the topic testing
-        with self.api.pubsub_sub(topic) as sub:
+        sub = self.api.pubsub_sub(topic)
 
+        channels = None
+        try:
             # grab the channels we're subscribed to
             channels = self.api.pubsub_ls()
+        finally:
+            sub.close()
 
-            assert channels == expected_return
+        assert channels == expected_return
 
     def test_pubsub_peers(self):
         """
--- a/test/functional/tests.py
+++ b/test/functional/tests.py
@@ -577,31 +577,31 @@ class IpfsApiMFSTest(unittest.TestCase):
             self.api.files_stat(self.test_directory_path)
 
 
-@skipIfOffline()
-class TestBlockFunctions(unittest.TestCase):
-    def setUp(self):
-        self.api = ipfsapi.Client()
-        self.multihash = 'QmYA2fn8cMbVWo4v95RwcwJVyQsNtnEwHerfWR8UNtEwoE'
-        self.content_size = 248
+# @skipIfOffline()
+# class TestBlockFunctions(unittest.TestCase):
+    # def setUp(self):
+        # self.api = ipfsapi.Client()
+        # self.multihash = 'QmYA2fn8cMbVWo4v95RwcwJVyQsNtnEwHerfWR8UNtEwoE'
+        # self.content_size = 248
 
-    def test_block_stat(self):
-        expected_keys = ['Key', 'Size']
-        res = self.api.block_stat(self.multihash)
-        for key in expected_keys:
-            self.assertTrue(key in res)
+    # def test_block_stat(self):
+        # expected_keys = ['Key', 'Size']
+        # res = self.api.block_stat(self.multihash)
+        # for key in expected_keys:
+            # self.assertTrue(key in res)
 
-    def test_block_get(self):
-        self.assertEqual(len(self.api.block_get(self.multihash)), self.content_size)
+    # def test_block_get(self):
+        # self.assertEqual(len(self.api.block_get(self.multihash)), self.content_size)
 
-    def test_block_put(self):
-        path = os.path.join(os.path.dirname(os.path.dirname(__file__)),
-                            "functional", "fake_dir", "fsdfgh")
-        expected_block_multihash = 'QmPevo2B1pwvDyuZyJbWVfhwkaGPee3f1kX36wFmqx1yna'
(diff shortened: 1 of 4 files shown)


test/functional/tests.py
```diff
@@ -807,6 +807,68 @@ def test_bitswap_unwant(self):
...
+    def setUp(self):
+        self.api = ipfsapi.Client()
+
+    def test_pubsub_pub_sub(self):
+        """
+        We should test both publishing and subscribing
+        at the same time for it's difficult
+        to determine if one works without checking the
+        other.
+        """
```
alexander255: This sentence confused me originally when reading it for the first time: At least to me it sounds like a TODO-entry. Is this supposed to be present-tense (ie: “We test both publishing and subscribing at the same time because …”) by any chance?

test/functional/tests.py
```diff
@@ -807,6 +807,68 @@ def test_bitswap_unwant(self):
...
+        # the topic that will be published/subscribed to
+        topic = 'testing'
```
alexander255: Please use an unique channel topic like per session (for instance by using uuid like in `"{}.testing".format(uuid.uuid4())`): As far as I know pubsub messages are supposed to be impliciently propagated across subscribing nodes in a future version and with these generic names there'd be a high chance of random interference from other nodes (and also interferance from other people possibly running these same unit tests) which would cause “random errors in tests” – something that everybody hates to debug.

(Also in the other test below.)

test/functional/tests.py
```diff
@@ -807,6 +807,68 @@ def test_bitswap_unwant(self):
...
+    def test_pubsub_ls(self):
+        """
+        Testing the ls, assumes we are able
+        to at least subscribe to a topic
+        """
+        topic = 'lsing'
+        expected_return = { 'Strings': [topic] }
+
+        # subscribe to the topic testing
+        # in this test we don't care about the return value
```
alexander255: Maybe add something like “… but we need to keep it around for the lifetime of this function to ensure that we stay subscribed to the topic long enough” to make it more obvious what is happening?

test/functional/tests.py
```diff
@@ -577,31 +577,31 @@ def test_dir_make_fill_list_delete(self):
...
+
+    # def test_block_put(self):
+        # path = os.path.join(os.path.dirname(os.path.dirname(__file__)),
+                            # "functional", "fake_dir", "fsdfgh")
+        # expected_block_multihash = 'QmPevo2B1pwvDyuZyJbWVfhwkaGPee3f1kX36wFmqx1yna'
+        # expected_keys = ['Key', 'Size']
+        # res = self.api.block_put(path)
+        # for key in expected_keys:
+            # self.assertTrue(key in res)
+        # self.assertEqual(res['Key'], expected_block_multihash)
```
alexander255: Why comment out the block test?

test/functional/tests.py
```diff
@@ -807,6 +807,68 @@ def test_bitswap_unwant(self):
         result = self.api.bitswap_unwant(key='QmZTR5bcpQD7cFgTorqxZDYaew1Wqgfbd2ud9QqGPAkK2V')
         self.assertTrue(result is not None)
 
+@skipIfOffline()
+class IpfsApiPubSubTest(unittest.TestCase):
+
+    def setUp(self):
+        self.api = ipfsapi.Client()
+
+    def test_pubsub_pub_sub(self):
```
alexander255: Any ideas why this test fails on the CI system? (Error is at https://travis-ci.org/ipfs/py-ipfs-api/jobs/371828905#L621.)

test/functional/tests.py
```diff
@@ -807,6 +807,68 @@ def test_bitswap_unwant(self):
...
+        # grab the channels we're subscribed to
+        channels = self.api.pubsub_ls()
+        self.assertEqual(channels, expected_return)
+
+    def test_pubsub_peers(self):
+        """
+        Not sure how to test this since it fully depends
+        on who we're connected to. We may not even have
+        any peers
+        """
```
alexander255: Indeed :slightly_smiling_face: 
But please call the function anyways and check if its result appears to make sense (a list with some values probably), so that we can at least detect if the function is totally broken for some reason.

alexander255:
> Hello everyone. Apologies if there are mistakes in this PR. First time contributing to a project.
>
> I have implemented python functions for all pubsub methods: ls, pub, sub, and peers. I have also included a few test cases (that all pass) for the aforementioned methods. Note that the test case for "peers" does not actually test anything for I can't predict what peers the function will return.

Thanks for providing these! They're very welcome!

> Another note for the sub api. I didn't want to introduce another dependency for a webhook (?) so I simply returned a stream for that function. In the documentation for the pubsub_sub function I made it clear if a user wants to obtain published content they simply need to iterate over the generator (of the stream) that's returned.

No problem! The only way to do fully concurrent requests is to launch another thread currently. You pretty much already got the best of the underlying `requests` library with your approach – anything more would require that port to `aiohttp` that I had started some time ago, but never completed due to some hard-to-debug issues arising that I couldn't figure out.

> Once again this is my first time contributing to an open source project so I apologize for any mistakes. I ran `tox` for both codestyle and test cases.

Again, thanks for the work you provided already! :+1:

test/functional/tests.py
alexander255: Actually the issue may simply be that the versions of `go-ipfs` used by the CI system is too old. Try updating the URL in https://github.com/ipfs/py-ipfs-api/blob/master/.travis.yml#L17 to whatever the URL of the latest Linux-x86_64 version is. The issue should likely be fixed by that.

test/functional/tests.py
```diff
@@ -807,6 +807,68 @@ def test_bitswap_unwant(self):
...
+        # make sure something was actually returned from the subscription
+        assert sub is not None
+
+        # publish a message to topic
+        self.api.pubsub_pub(topic, message)
+
+        # get the 
+        sub_data = next(sub)
+
+        self.assertEqual(sub_data['data'], expected_data)
```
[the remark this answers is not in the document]
slandau3: Didn't know that. Will do!

test/functional/tests.py
slandau3: I commented that out when I was testing on my machine. For some reason it was taking a really long time and I forgot to uncomment it. My bad.

slandau3:
Thanks for taking a look at it :smile: ! I'll make the changes you recommended as soon as I get a chance

test/functional/tests.py
slandau3: This reminds me, when requests performs a request to the subscription url, I believe it will remain open. This could waste resources if someone is constantly subscribing and discarding channels. I don't think it'll close itself because it's still technically listening to the channel. What do you recommend doing for this? Wrapping it in a class that has a function which will close the subscription?

test/functional/tests.py
whereswaldon: We've definitely seen instances of that in the past, so I agree that this is likely the issue.

whereswaldon:
Thanks for all of your work so far! I agree with @alexander255's comments. This is close to merge-worthy though. Well done!

slandau3:
Thank you so much @whereswaldon and @alexander255! It's been my pleasure and I'm really glad you guys like it! I really like the idea of IPFS and I want to be able to assist in it's development and adoption in any way possible.

All the tests I wrote appear to pass :smile: yet several tests I did not touch are failing. Those tests also fail on the master branch itself

test/functional/tests.py
slandau3: Added a call to the function as well as a few assertions

whereswaldon:
I decided to resolve the travis merge conflict in favor of 0.4.14 because it's the latest stable release. It seems like it might be a bad idea to run our tests against a release candidate version.

slandau3:
Awesome. All tests pass. Is there anything left to do before merge? I believe all the changes @alexander255 requested have been made

alexander255:
Its kind of unfair that we're unloading our technical depth on your PR @slandau3, but at least some issues are now being ironed out thanks to you. Aside from documenting the need to close the returned generator I have no further objects.

slandau3:
That's quite all right @alexander255! I appreciate it because it helps me get a better understanding of the program. Glad it's resulted in other issues being ironed out too.

I just updated the tests to close the "generator" returned. I also updated the documentation for pubsub_sub to make it clear to the reader that the returned subscription is a context manager/can be manually closed with `.close()`

I'm happy to help!

slandau3:
I'm not sure what the problem is with test_block_get or those metric tests. I don't know if anything I did could have affected them... I doubt it because I only made minor changes since the last time it said "all tests pass"

whereswaldon:
I restarted the Python 2.7 build and it succeeded. Seems like we have some nondeterminism in the tests, which is alarming.

alexander255:
The (occasional) errors with `test_block_get` are annoying and indeed non-deterministic, but they do not represent any issues that need fixing as such (at least on our end): They're simply artifacts caused by the shitty network environment for P2P applications in general – strict NAT with no possibility of port forwarding – inside the CI system. Of course, go-IPFS should probably somehow handle this situation better, but it's tricky and the options are very limited. I hope this will be improved cause many people are stuck in this kind a network environment, but it's not something we can fix within the scope of this project.

slandau3:
Just finished making the changes that were requested

slandau3:
Seems I forgot to add the updates client.py file in the latest commit which is why it's failing some tests (for ls). I'll add it once I get home today

alexander255:
@slandau3: I like the idea of exposing a more problem-specific interface for subscription channels as calling `next(sub)` to get a new message feels kinnda weird. :+1:
As you probably noticed,  it's still missing the `.close` method however. You can also drop the `.__next__` method as well and make `.__iter__` just `return self.__sub` – it's shorter and this way `for msg in sub: …` will work again (the current implementation does not return an iterator).

slandau3:
@alexander255 Sorry for the wait. I made the changes you requested and I'm not sure what I was thinking with the original `__iter__`. Not sure why I used `return` instead of `yield` :thinking:

slandau3:
Thanks @alexander255! Glad I could help. I'll keep a lookout for other features I can work on in the future.

PR #124, opened 2018-05-01
Update CI to go-IPFS v0.4.14

alexander255:
Fixes the CI issues seen in #123.

### DIFF
--- a/ipfsapi/client.py
+++ b/ipfsapi/client.py
@@ -2056,7 +2056,12 @@ class Client(object):
         ``ipfsapi.exceptions.ConnectionError``, until you start another IPFS
         daemon instance.
         """
-        return self._client.request('/shutdown')
+        try:
+            return self._client.request('/shutdown')
+        except exceptions.ConnectionError:
+            # Sometimes the daemon kills the connection before sending a
+            # response causing an incorrect `ConnectionError` to bubble
+            pass
 
     ###########
     # HELPERS #
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
     before_install: true
 
 before_install:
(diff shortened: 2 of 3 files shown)


ipfsapi/client.py
```diff
@@ -2056,7 +2056,12 @@ def shutdown(self):
         ``ipfsapi.exceptions.ConnectionError``, until you start another IPFS
         daemon instance.
         """
-        return self._client.request('/shutdown')
+        try:
+            return self._client.request('/shutdown')
+        except exceptions.ConnectionError:
+            # Sometimes the daemon kills the connection before sending a
+            # response causing an incorrect `ConnectionError` to bubble
```
whereswaldon: Hmmm... should we file that as a bug with go-ipfs?

whereswaldon:
It seems to me that we spend a lot of time fixing tests that break because old assumptions are no longer true. For instance, whenever they add a key or change the order of a result, we usually start failing some tests. Do you think that's a methodological problem on our part?

ipfsapi/client.py
alexander255: Indeed! Filed it here: https://github.com/ipfs/go-ipfs/issues/4994

alexander255:
I agree that it's annoying, but, in a way, we *want* this kind of breakage so that we notice when something changes and can add the new data to the tests. This is good because, from that point on, the additionally returned data becomes part of the API and if the extra fields change or are removed after this then that'd be a regression.

– Of course all of this is indeed useless if we don't actually take it seriously when it happens, so I filed https://github.com/ipfs/go-ipfs/issues/4995 about the now-missing `'Links'` datastructure in `object_put`. :relaxed:

PR #125, opened 2018-05-02
Allow generators returned by `stream=True` commands to be closed

alexander255:
This adds documentation for the calls that already use `stream=True` as well as making it possible to use the returned generator in a context manager (`with`-statment) for easier cleanup.


ipfsapi/http.py
```diff
@@ -38,6 +38,56 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+def stream_decode_iter(response, parser):
+    try:
+        for data in response.iter_content(chunk_size=None):
+            #PY2: Use `yield from` instead
```
whereswaldon: So you're saying that "if we didn't have to support python2, we'd use that syntax"?

alexander255:
Adding (unit) tests is not silly at all and doing that actually made me find a relatively major issue: The response wouldn't be closed if it hadn't been iterated at least once (because the iterator is not started until you call `next` on it at least once, so the `try`…`finally` statement was never entered). I rewrote most of my patch because of this, but it should now be better and tested now.

PR #128, opened 2018-07-08
add 'offset' and 'length' parameters to cat method

cshuaimin:
It's useful to get one block of a file, such as using libmagic to determine the file type.

### DIFF
--- a/ipfsapi/client.py
+++ b/ipfsapi/client.py
@@ -2378,7 +2378,7 @@ class Client(object):
             ... c.pubsub_pub('testing', 'hello')
             ... for message in sub:
             ...     print(message)
-            ...     # Stop reading the subscription after 
+            ...     # Stop reading the subscription after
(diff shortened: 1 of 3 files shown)


alexander255:
Thanks for this! Looking good!

Please add some tests however as we're aiming to increase our test coverage.

cshuaimin:
@alexander255 OK.

The `ipfsapi.http.stream_decode_full()` function will return an empty list `[]` if I call `api.cat(<hash>, offset=0, length=0)`, which should be an empty `bytes` `b''`. There should be a better way to fix this.

https://github.com/cshuaimin/py-ipfs-api/blob/1fb5b8077790a396ba647b7c8635989a2aa4e40f/ipfsapi/http.py#L116-L117

alexander255:
Looking good! I see no issues in merging this now. :slightly_smiling_face: 

Thanks for your contribution!

PR #139, opened 2018-11-02
Fix a Bug While Upload directories in Windows

IvanLuLyf:
When add a directory in Windows System,the result always return with ```QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn```.
Because the dir separator using '\\' in Windows but needed is '/'.
I fix this bug for Python's ipfs api.


alexander255:
LGTM, Thanks!

PR #145, opened 2018-11-23
Rewrite `ipfsapi/multipart.py` to address the issues mentioned in #104

### LINKED ISSUE #104: `.add()` serializes all input data to a giant binary string before uploading
This is obviously a no-go.

alexander255:
Fixes #104 and some other stuff. Also gets rid of `requests` in the multipart code making it framework agnostic again, while probably being more efficient, but definitely more resource aware:

  * Even when streaming large directory trees the maximum number of open fds should only be `tree_depth + 1` (one fd for each directory layer and one for the actual file in question)
  * Files are streamed from disk using generators so there should never be more than a couple of kilobytes of file data be present in RAM

### DIFF
--- a/ipfsapi/multipart.py
+++ b/ipfsapi/multipart.py
@@ -2,691 +2,713 @@
 """
 from __future__ import absolute_import
 
-import re
-import requests
-import io
+import inspect
 import os
-from inspect import isgenerator
-from uuid import uuid4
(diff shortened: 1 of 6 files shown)


alexander255:
Fixed Python-3.4 (only) compat issue

PR #149, opened 2018-12-18
Upgrade API to comply to the IPFS Interface specification (py-ipfs-http-client version)

alexander255:
Follow-up from #146. Needs discussion on that PR to complete before it can be merged.


alexander255:
Finally!

PR #153, opened 2019-01-29
Improve testing

alexander255:
Add testing on macOS and much improved test coverage data collection (which previously was broken)


alexander255:
With build being at a crawl right now this could take a while…

PR #155, opened 2019-01-31
change pass_defaults to do a deep merge, enabling HTTP basic auth

### LINKED ISSUE #154: Basic auth support
For people using a private IPFS cluster with a swarm key, it is also useful to have support for HTTP basic auth. Interaction with the gateway is easy, as there is no library and you can just use URLs like `https://user:pass@gatewayurl.com/ipfs/hash`. However, for using the IPFS API, this is not as simple.

Making `py-ipfs-api` support basic auth and is a useful and simple change that I'm willing to implement. Instead of vendoring this code I would love to submit a PR. Just want to ask before doing this - would you accept a PR for basic auth support?

ldub:
Fixes #154 

Enables basic auth and many other use cases by "deep merging" the defaults dict instead of using `dict.update` which overwrites previously seen keys (such as `headers`).

I do see that @alexander255 and @AliabbasMerchant are doing a lot of new work and reorganization on the py-ipfs-http-client branch. This change would be very simple to add to that branch as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,10 +143,14 @@ This module also contains some helper functions for adding strings and JSON to I
 Use an IPFS server with basic auth (replace username and password with real creds):
 
 ```py
->>> import base64
 >>> import ipfshttpclient
->>> creds = base64.b64encode(b"username:password").decode("utf-8")
->>> headers = {"Authorization" : "Basic " + creds}
+>>> api = ipfshttpclient.connect('/dns/ipfs-api.example.com/tcp/443/https', username="foo", password="bar")
+```
+
+Pass custom headers to your IPFS api with each request:
+```py
+>>> import ipfshttpclient
+>>> headers = {"CustomHeader": "foobar"}
 >>> api = ipfshttpclient.connect('/dns/ipfs-api.example.com/tcp/443/https', headers=headers)
 ```
 
--- a/README.md
+++ b/README.md
@@ -140,6 +140,18 @@ This module also contains some helper functions for adding strings and JSON to I
 [1, 77, 'lol']
 ```
 
+Use an IPFS server with basic auth (replace username and password with real creds):
+
+```py
+>>> import base64
+>>> import ipfsapi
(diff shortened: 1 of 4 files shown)


ldub:
Build is failing on codestyle test that is not related to this change.

alexander255:
Looks good! But having to manually build the `Authentication` header feels quite clunky from a user perspective to me. Could you maybe add some dedicated `username` and `password` parameters to the `connect` function and `Client` constructor and have them pass those values down to the HTTP library (which _still_ is `requests`)?
Also, please target the `py-ipfs-http-client` branch, `master` will not receive any further changes until we're done with the transition.

Thanks for your help!

ldub:
Had a bit of git trouble rebasing onto the newly transitions `ipfshttclient` code, but looks fine now. I haven't implemented the `username` & `password` parameters yet but will do this soon.

alexander255:
Just leave a comment when you feel its ready, will review then.

ldub:
Hi @alexander255 sorry for the delay. I've added the `username` and `password` arguments into the top level `ipfshttpapi.connect()` method, but instead of passing them down to `Client`, `requests_wrapper`, and `request`, I just convert them into the basic auth header at the top level and pass that header down. Let me know if this is OK.

I tested this on my own IPFS server behind basic auth, I am happy to PM you the credentials if you'd like to test it as well: https://keybase.io/dubinets or your preferred chat platform.

alexander255:
Sorry, this obviously got lost in my mail. While the code you wrote looks reasonable, the auth credentials need to be passed down to the `.http.HTTPClient` constructor, because that serves as our abstraction of “a transport for sending requests to an IPFS daemon”. The upper layer otoh then provides wrappers for individual API requests and is independent of the HTTP protocol. Also we seek to provide the same interface on `.Client` vs `.connect` in all possible cases.
Ideally please just pass the auth information all the way down to requests, so that we don't duplicate any functionality that the HTTP library already implements.

Thanks for still working on this!

Issue #150, opened 2019-01-25
Default values of `Client`/`connect()` are not documented


AuHau:
The #7 was reintroduced I guess by some refactoring. Passing None to `Client`/`connect()` will result in no default values used and a not much informative exception:
```
File "/Users/c5272397/dev/ipfs-publish/.direnv/python-3.6.5/lib/python3.6/site-packages/ipfsapi/http.py", line 143, in __init__
    if not re.match('^https?://', host.lower()):
AttributeError: 'NoneType' object has no attribute 'lower'
```

I am happy to provide refix, I am only not sure how to proceed because of the on-going renaming (#148). Should I wait until merged, target some specific branch or something else? Please let me know.

AuHau:
Okey! I will wait and follow the #149 and after completion then revisit.

Thanks for the reply!

alexander255:
As mentioned in #150 this won't be “fixed” by allowing `None`; rather we'll update the documentation to highlight that `ipfshttpclient.DEFAULT_{HOST,PORT,BASE}` are the respective default values.

PR #156, opened 2019-01-31
(#150) Allow None for connect()/Client to fallback to default values

AuHau:


AuHau:
Closing in favor of #162

PR #159, opened 2019-02-03
client: add support for nocopy, raw-leaves to add method

radfish:
nocopy is accessible form the CLI, so let's make it accessible from API as well.

### DIFF
--- a/test/functional/test_files.py
+++ b/test/functional/test_files.py
@@ -185,6 +185,7 @@ def test_add_relative_path(client, cleanup_pins):
 	assert res["Hash"] in client.pin.ls(type="recursive")["Keys"]
 
 
+@pytest.mark.skip(reason="Requires go-ipfs 77cd41acfef6cc442309eda565c29e36fdd488d9 PR#5937")
 def test_add_nocopy_with_relative_path(client):
 	error_msg = None
 	try:
--- a/test/functional/test_files.py
+++ b/test/functional/test_files.py
@@ -16,6 +16,11 @@ FAKE_FILE2_PATH = conftest.TEST_DIR / "fake_dir" / "popoiopiu"
 
 FAKE_FILE1_HASH = {"Hash": "QmQcCtMgLVwvMQGu6mvsRYLjwqrZJcYtH4mboM9urWW9vX",
                    "Name": "fsdfgh", "Size": "16"}
+FAKE_FILE1_RAW_LEAVES_HASH = {
+	"Hash": "zb2rhXxZH5PFgCwBAm7xQMoBa6QWqytN8NPvXK7Qc9McDz9zJ",
+	"Name": "fsdfgh", "Size": "8"
+}
+
 FAKE_FILE1_DIR_HASH = [
 	{"Hash": "QmQcCtMgLVwvMQGu6mvsRYLjwqrZJcYtH4mboM9urWW9vX",
 	 "Name": "fsdfgh", "Size": "16"},
@@ -103,6 +108,13 @@ FAKE_DIR_HASH = [
 ]
 
 
+def calc_path_rel_to_cwd(p):
+	p = str(p)  # for Python < 3.5
+	prefix = os.path.commonprefix([p, os.getcwd()])
+	relpath = os.path.relpath(p, prefix)
+	assert not os.path.isabs(relpath)
+	return relpath
+
 
 def test_add_single_from_str_with_dir(client, cleanup_pins):
 	res = client.add(FAKE_FILE1_PATH, wrap_with_directory=True)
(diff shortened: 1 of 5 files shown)


AuHau:
I believe that the target branch should be `py-ipfs-http-client` and not `master`.
The `py-ipfs-http-client` will soon replace master.

radfish:
Ok. Rebased and re-tested.

alexander255:
Looks good and I'd love to just merge this. However… Two things:

  * As far as I know, until recently the `nocopy` parameter did not work correctly when using the API since the API would stream a copy of the local file using HTTP to the daemon which would then blindly add it to its block store – `nocopy` flag or not. Could you please check that this works with the [recent fix in `go-ipfs`](https://github.com/ipfs/go-ipfs/issues/4558)? The linked issues has good “steps to reproduce”, just use the Python API (`ipfshttpclient.connect().add("filepath", nocopy=True)`), rather then `curl` for uploading.
  * And while we're at testing: Please also add a test that uses this new parameter to `test/functional/files.py`. Ideally the test would be similar to what was described in the linked issue but a basic test will do too.

radfish:
1. Confirmed, tested on go-ipfs 77cd41a.  I did not notice that blocks were not actually added to the filestore, thanks for pointing it out. This time I checked with `filestore ls` and `dag get`, but see notes on testing nocopy below. This PR adds the Abspath header for absolute paths -- see rationale in commit message below.

2. Added tests for the new arguments. Tests testing error path depend on the error report added to the server in 77cd41a, so perhaps can't be released until go-ipfs is released with that fix.

Note that server is still broken: it does not return an error if the path is invalid, it again silently ignores nocopy and adds to blockstore in this case. Also note that, relative paths are invalid (neither relative to server user home nor relative to working dir of server is works), despite paths in the filestore being recorded relative to the user home.

Finally note that adding a file with nocopy whose blocks happen to already be in the blockstore, is going to succeed, but those blocks won't be (also) added to the filestore, they won't be listed as duplicates by `filestore dups`. So, if you test nocopy but don't see the file in the filestore, that may be normal. (I don't find this behavior intuitive for getting your data into the filestore, so I patched my IPFS filestore/filestore.go Has method to allow duplicate adds to the filestore, but also need to patch the `block rm` command to support removing from blockstore only, in order to be able to remove the data in duplicated in the blockstore.).

```
commit 530b8bc99153b5567422f37cfc4ec26b13a6f0c9 (HEAD -> PR--add-nocopy, rf/PR--add-nocopy)
Author: redfish <redfish@galactica.pw>
Date:   Sun Feb 3 13:51:22 2019 -0500

    client: add nocopy, raw-leaves options to add cmd
    
    The nocopy command is for adding files by reference to their
    location in the filesystem. The nocopy implies raw-leaves.
    
    Requires bug fix in go-ipfs 77cd41a (see issue #4558), ver >0.4.18
    (strictly greater).
    
    The nocopy option implies raw-leaves, and the default for raw-leaves
    depends on nocopy. This dynamic default is the behavior of the HTTP API
    tested via curl.
    
    Adding files to the filestore via nocopy requires that the path
    is absolute, not relative. The multipart stream component appens
    the Abspath header when it has the absolute path to the file.
    
    In IPFS daemon used for tests, enable the filestore feature.
    
    The tests could leverage the filestore API, but that part of the API is
    not yet implemented, so leaving parts of the test commented out.

commit db87134d4ba02ee955464055a9eb4f81ff450350
Author: redfish <redfish@galactica.pw>
Date:   Sat Feb 9 20:22:54 2019 -0500

    multipart: add Abspath header when have abs path
    
    This header is required for adding files by reference, i.e.  adding into
    the filestore via the add command with nocopy=True option.
    
    We can't unconditionally add the header, because we don't always have
    the path to the file (e.g. not when the caller of add() passes us a file
    descriptor only). Unconditionally adding a header with a blank value
    does not seem good. We only append the Abspath header if we have a path
    and if it is an absolute path.  Note that Abspath value must be an
    absolute path -- the nocopy option only makes sense with an absolute
    path.
    
    Regardless of this client side decision for when to append the header,
    though, the server should return an error if a relative path or a
    non-existant path is passed in Abspath. At the very least, it should
    return an error nocopy is passed with a relative path or non-existant
    path or a path outside of the daemon's working directory.  Currently,
    the server does NOT return an error, instead it silently ignores nocopy
    and adds the file to the blockstore instead -- this is a bug.
    
    Luckily, the server does return an error if nocopy is passed but Abspath
    header was not set. The client-side logic described above reduces the
    case of nocopy with relative path to nocopy without Abspath header, and
    thus the caller will get an error in all cases.

commit 4db4de7a79cd5ffd9c1215713cebba66bac12ab9
Author: redfish <redfish@galactica.pw>
Date:   Sat Feb 9 23:20:16 2019 -0500

    multipart: fix warning about regexp flags
    
    ./ipfshttpclient/multipart.py:362: DeprecationWarning: Flags not at
    the start of the expression '^.*\\Z(?ms)$'

commit 2cc4d051c6d3c858059178335446b5905f788a4e
Author: redfish <redfish@galactica.pw>
Date:   Sat Feb 9 23:23:16 2019 -0500

    test: unit: multipart: use regexp literals
    
    Use r'' when defining literals that form a regular expression.
    Fixes warning about invalid escape characters, like \S.
```

test/functional/test_files.py
```diff
@@ -126,11 +135,57 @@ def test_only_hash_file(client):
 	assert res["Hash"] not in client.pin.ls(type="recursive")
 	assert res["Hash"] not in list(map(lambda i: i["Ref"], client.unstable.refs.local()))
 
-
 def test_add_multiple_from_list(client, cleanup_pins):
 	res = client.add([FAKE_FILE1_PATH, FAKE_FILE2_PATH])
 	assert FAKE_FILES_HASH == res
 
+def test_add_with_raw_leaves(client):
```
alexander255: Please add the `cleanup_pins` fixture to all of your tests, like so:

```py
def test_add_with_raw_leaves(client, cleanup_pins):
	# Body
```

This way the pins implicitly added will be removed again after the test.

test/unit/test_multipart.py
```diff
@@ -192,35 +194,54 @@ class TestStreamFileMixin(unittest.TestCase):
...
 		file.write(b"!234")
 		file.seek(0)
 
-		expected = b'--' + generator._boundary.encode() + b'\r\nContent-Disposition: file; '\
-		         + b'filename="functional%2Ffake_dir%2Ffsdfgh"\r\nContent-Type: '\
-		         + b'text/plain\r\n\r\n' \
+		expected = b'--' + generator._boundary.encode() + b'\r\n'
+		expected += b'Abspath: ' + name.encode(ENC) + b'\r\n' if abspath else b''
+		expected += b'Content-Disposition: file; '\
+		         + b'filename="' + urllib.parse.quote_plus(name).encode(ENC) + b'"\r\n'\
```
alexander255: Replace `urllib.parse.quote_plus(name).encode(ENC)` by `urllib_parse.quote_plus(name).encode(ENC)` and add the import `from six.moves import urllib_parse` to the top of the file to make the Python 2.7 error go away.

test/functional/test_files.py
```diff
@@ -103,6 +107,11 @@
 ]
 
 
+def calc_path_rel_to_cwd(p):
+	prefix = os.path.commonprefix([p, os.getcwd()])
+	relpath = os.path.relpath(p, prefix)
```
alexander255: Use `str(p)` instead of `p` here to make the Python 3.5- error go away.

test/functional/test_files.py
radfish: Added fixture to non-error test cases.

test/unit/test_multipart.py
radfish: Fixed.

test/functional/test_files.py
radfish: Fixed.

radfish:
Ok, applied all the fixes. Thanks for specific patches.
CI timed out on the stat test.

alexander255:
@radfish: Thank you! This looks perfect! :grin:

PR #160, opened 2019-02-04
Allow specifying return_result=False to speed up gc()

### LINKED ISSUE #82: Add HEAD option / ignore result for some requests
For some requests it makes sense to use HEAD over GET if it is only
relevant whether the request succeeded or not.

Background:

`client.repo_gc()` can potentially return many values (scale: easily 10-100k) depending on how much garbage was collected.
It takes quite some time to parse and decode (`json`) all that even if the result does not matter
inside the calling function.
Most of the time you do not need to know the hashes of the garbage.

lordcirth:
gc() can now use HTTP's HEAD instead of GET, thereby not returning the list of all hashes that were removed.

Resolves #82.


alexander255:
I saw you commit yes, but I wasn't able to look at it yet.

Regarding this: I feel a bit uncomfortable exposing such a low-level detail (the HTTP request method) like this, but this is definitely useful nonetheless. Making the parameter name more descriptive would be an improvement, I think. How about calling it `return_result=False` or similar?

Some notes about this:

  * The parameter should only change the request method if no specific request method is required by the backend; so `return_result=False` on `.add()` should have no effect on the HTTP method used since it needs to be `POST` in that case afaik
  * When the library users says they expect no result, then the `stream` parameter should be ignored
  * …and the backend should mandatorily return `None` (no result) even if the daemon did not

Also: Do all IPFS endpoints support the `HEAD` method (that's what I'm assuming here) or just this one? Do you happen to know which version introduced this feature?

Thank you for helping out with this!

lordcirth:
Yeah, so I'm not entirely happy with exposing 'method' either.  I actually first wrote 'head=True', but the problem is that gets passed through **kwargs, and then the request breaks.  I'm not very experienced with Python. Could we just stop passing **kwargs to request?

alexander255:
@lordcirth: I'm not sure what you're getting at? Why not simply add a method named `result` instead of the current `method` parameter?

The processing logic would then be something like:

```py
if files or data:
    method = "POST"
elif not result:
    method = "HEAD"
else:
    method = "GET"

if not result:
    decoder = None
    stream = False
```

Also the parsed result should not be returned, but immediately discarded.

What do you mean _stop passing **kwargs to request_?

lordcirth:
> What do you mean _stop passing **kwargs to request_?

In my first draft, I added to gc() an boolean argument 'head', which defaulted to False. If set to True, it would use HTTP HEAD instead. But (and I am not good at Python, so I might be wrong) in

`return self._client.request('/repo/stat', decoder='json', **kwargs)`,

`**kwargs` passes every keyword argument that `gc()` received, to `request()`. `request()` then threw an error because `head` was not a valid argument.

alexander255:
Hmm… I guess that may be part of the communication problem here. I was talking about adding the `result` parameter for all client API methods (like you did with the `method` parameter). Basically keep everything the way you figured it out, but expose `result` rather than `method` and determine the HTTP method to use from that (using the code I provided in my last message).

In other words: The way things currently are is fine architecturally, I just want you to expose a parameter with more semantic meaning then the raw HTTP request method.

lordcirth:
I see my mistake. I was confusing _client.request(),  _client._request(), and so on.

lordcirth:
Ok, can you review it again? Thanks!

lordcirth:
So, I rebased, and now I'm getting tons of lint errors on commit:

```
./ipfshttpclient/client/base.py:20:30: E127 (flake8-tabs) unexpected number of tabs and spaces within expression (expected 4 tabs and 0 spaces, got 3 tabs and 26 spaces)
...
```
Seems there's mixed tabs and spaces. I noticed before that some files use tabs and some spaces, it's really annoying. What are we supposed to be using?

radfish:
> Seems there's mixed tabs and spaces. I noticed before that some files use tabs and some spaces, it's really annoying. What are we supposed to be using?

As far as I found out: tabs for indentation but spaces for alignment, so as many tabs as required by the level, and then add spaces if you want to align cosmetically.

(But, some files appear to use spaces exclusively).

Btw, I also had to periodically `rm -rf .tox` in order to force the checker to update.

lordcirth:
> > Seems there's mixed tabs and spaces. I noticed before that some files use tabs and some spaces, it's really annoying. What are we supposed to be using?
> 
> As far as I found out: tabs for indentation but spaces for alignment, so as many tabs as required by the level, and then add spaces if you want to align cosmetically.
> 
> (But, some files appear to use spaces exclusively).
> 
> Btw, I also had to periodically `rm -rf .tox` in order to force the checker to update.

http.py has mixed tabs and spaces already. It's quite annoying, but it seems that tox will permit it...

lordcirth:
@alexander255 Ok, it's been a while, but you can you please review again? Thanks.

alexander255:
Sorry, was busy.

Changes LGTM now. Thank you for not giving up. :slightly_smiling_face: 

> http.py has mixed tabs and spaces already. It's quite annoying, but it seems that tox will permit it...

Will be fixed soon™

PR #162, opened 2019-02-09
Improve handling of default values

AuHau:


alexander255:
Thank you for pointing out this deficiency! I will not merge this code however, since I disagree with you on `None` being a sensible value for the host, port or base path fields. Instead I'll see to it, that the documentation is updated to point out that the defaults for these fields are `ipfshttpclient.DEFAULT_{HOST,PORT,BASE}` respectively.
Thanks for proposing this however.

PR #163, opened 2019-02-09
Improve documentation on ttl option

AuHau:
See: https://github.com/ipfs/go-ipfs/pull/5979


alexander255:
Thanks! Always nice to have documentation improvements as well.

(The timeout isn't your fault, so merging.)

PR #169, opened 2019-02-16
Add timeout option

AliabbasMerchant:
Fixes #129   
@alexander255 Please close #143   

Also, most probably a problem with my system, but _all_ the tests are failing. I am getting a `UnicodeDecodeError`.  
Please check once before merging.

PR #170, opened 2019-02-21
Add timeout option

alexander255:
Follow-up of #169 using `pytest-localserver` instead of `httmock` for testing.


alexander255:
I gave in on the networked part of the block tests and changed them in a way to ensure that each fetched block is put first. CI times are now a lot lower and we don't get those stupid timeouts anymore.

PR #172, opened 2019-03-02
name tests and DHT-related arguments for publish

radfish:
Tests take a long time to run. Would be nice if daemon supported offline mode; or at least support local resolve to resolve own keys instantly.

### DIFF
--- a/ipfshttpclient/client/name.py
+++ b/ipfshttpclient/client/name.py
@@ -6,7 +6,8 @@ from . import base
 
 class Section(base.SectionBase):
 	@base.returns_single_item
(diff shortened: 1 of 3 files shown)


radfish:
I finally got all tests to pass except MacOS. Dunno why the server there returns error when trying to resolve own key. Is it the same go-ipfs version as on the other build bots?

alexander255:
Thank you! Looking very good; the macOS thing looks like bad luck, since there shouldn't be any idfference and it did work with Py3.4. Hopefully things will get better with 4.19 in terms of speed.

PR #173, opened 2019-03-03
pin: add optional paths arg to ls

radfish:
A must have for using pin.ls on large repositories.

### DIFF
--- a/ipfshttpclient/client/pin.py
+++ b/ipfshttpclient/client/pin.py
@@ -36,7 +36,7 @@ class Section(base.SectionBase):
 	
 	
 	@base.returns_single_item
-	def ls(self, type="all", **kwargs):
+	def ls(self, *cids, **kwargs):
 		"""Lists objects pinned to local storage.
 
 		By default, all pinned objects are returned, but the ``type`` flag or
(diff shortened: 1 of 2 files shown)


test/functional/test_pin.py
```diff
@@ -18,6 +21,42 @@ def resources(client):
...
 
+def path_to_hash(client, path):
+    return path.split("/")[-1]
+
+
+def is_pinned(client, path):
+	error_msg = None
+	try:
+	    resp = client.pin.ls(path)
+	    assert path_to_hash(path) in resp["Keys"]
```
alexander255: There is probably no point in having the `path_to_hash` function separate here.

ipfshttpclient/client/pin.py
```diff
@@ -55,6 +55,8 @@ def ls(self, type="all", **kwargs):
 
 		Parameters
 		----------
+		path : str
+			The path(s) to object(s) to be listed
```
alexander255: This should be named `cids` like everywhere else (also document the plural parameter, since there is no mandatory first `cid` parameter). Also change the description to something like “Pinned IPFS objects to search for; if any are given, only the requested CIDs will be returned if they are, in fact, pinned – adding CID that is not pinned will result in an error”

ipfshttpclient/client/pin.py
```diff
@@ -67,11 +69,14 @@ def ls(self, type="all", **kwargs):
 		-------
 			dict : Hashes of pinned IPFS objects and why they are pinned
```
alexander255: Add the following above the *Returns* section to document the exception raised:

```
		Raises
		------
		~ipfsapi.exceptions.ErrorResponse
			A CID was given, that was not actually pinned
```

ipfshttpclient/client/pin.py
```diff
@@ -55,6 +55,8 @@ def ls(self, type="all", **kwargs):
```
alexander255: Please extend this `code-block` with an example of `.pin.ls()` with parameters that succeeds and one that doesn't.

ipfshttpclient/client/pin.py
```diff
@@ -67,11 +69,14 @@ def ls(self, type="all", **kwargs):
 		-------
 			dict : Hashes of pinned IPFS objects and why they are pinned
 		"""
-		kwargs.setdefault("opts", {})["type"] = type
+		opts = {
```
alexander255: Add this comment above: `#PY2: No support for kw-only parameters after glob parameters`

test/functional/test_pin.py
radfish: removed

ipfshttpclient/client/pin.py
radfish: Renamed, re-wrote description. Btw, it's still 'path' in add,rm,verify.

ipfshttpclient/client/pin.py
radfish: Added.

ipfshttpclient/client/pin.py
radfish: Added.

ipfshttpclient/client/pin.py
radfish: Added.

ipfshttpclient/client/pin.py
radfish: Please unresolve if I didn't get the docs right.

alexander255:
Thank you!

PR #174, opened 2019-03-03
Offline mode

radfish:
My earlier prayer has been answered incredibly fast: go-ipfs 0.4.19 added an offline flag. Oh if only I had this yesterday... would have saved me hours. Now name tests run within a few seconds instead of tens of minutes!

I also tried launching the whole daemon with `--offline` but the bitswap and pubsub tests require online mode. Regardless of testing, though, the API should support offline at the granularity of connection and request, which this PR adds.

f96aaee is not strictly necessary, but it fixes the test hang when run with `--offline` daemon.

d43208d is unrelated. I just noticed the kwargs in the docs, but it was not in the code. It looked harmless and useful to add it, but if not, I can remove this commit from this PR.

### DIFF
--- a/ipfshttpclient/http.py
+++ b/ipfshttpclient/http.py
@@ -216,7 +216,8 @@ class HTTPClient(object):
     @pass_defaults
     def request(self, path,
                 args=[], files=[], opts={}, stream=False,
-                decoder=None, headers={}, data=None, timeout=120):
+                decoder=None, headers={}, data=None,
+                timeout=120, offline=False):
         """Makes an HTTP request to the IPFS daemon.
 
         This function returns the contents of the HTTP response from the IPFS
@@ -247,6 +248,9 @@ class HTTPClient(object):
             before giving up
             
             Defaults to 120
+        offline : bool
+                Execute request in offline mode, i.e. locally without accessing
+                the network.
         kwargs : dict
             Additional arguments to pass to :mod:`requests`
         """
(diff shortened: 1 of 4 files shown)


radfish:
I removed the commit that unskipped the skipped test. It's time to unskip it, but we can wait until build bots are updated to latest daemon. Also, until that bots are updated to 0.4.19, the timeout failures are back, partly made more likelky since I removed the timeout overrides from the name tests in this PR.

When should we expect build bots to be updated to 0.4.19?

alexander255:
Please drop these 2 commits:

  * client: base: document defaults parameter
  * http: accept kwargs in request method

The defaults parameter is considered an implementation detail since its behaviour is not well defined and somewhat unintuitive. `kwargs` was dropped because we consider our dependency on `requests` an implementation detail that is subject to change, I did notice however that the documentation was never updated and that the parameter never was removed from `HTTPClient.download`…

Other than I'm pretty stoked by the idea of finally having fast tests for good after having recently sped up the block tests by sometimes several minutes myself.

radfish:
Got it. Dropped.
Removed the kwargs doc entry in 8e1ec25.
Did not touch kwargs in download(): safe to remove? or backcompatibility concern? In any case, would not be part of this PR.

alexander255:
Thanks!

> Did not touch kwargs in download(): safe to remove? or backcompatibility concern?

Safe to remove; I just didn't know it was there. :slightly_smiling_face:
I'll remove it and also add documentation for the universal `offline` attribute.

PR #175, opened 2019-03-04
files.cp: fix return attribute in cp to be void

radfish:
Fixes assert failure introduced in 52df768c.
Added test that would have caught the problem.

```
AssertionError: Called IPFS HTTP-Client function should only ever return a list, when not streaming a response
```


alexander255:
Thank you! Looking good. :+1:

PR #194, opened 2019-11-18
bug fix: when storing large amounts of text or bytes the data would b…

WouterGlorieux:
…e truncated

Problem was because in the generator StreamBase._gen_chunks() the second parameter in the data slice was not incremented with the offset.
Because of this only the first chunk was yielded correctly.

### DIFF
--- a/ipfshttpclient/multipart.py
+++ b/ipfshttpclient/multipart.py
@@ -185,7 +185,7 @@ class StreamBase(object):
 		for data in gen:
 			#PERF: This is zero-copy if `len(data) <= self.chunk_size`
 			for offset in range(0, len(data), self.chunk_size):
-				yield data[offset:self.chunk_size]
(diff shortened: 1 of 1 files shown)


AliabbasMerchant:
I think @WouterGlorieux , you will have to change the base branch to master. @alexander255 am I right?

alexander255:
Damn, this is so stupid!

Thanks alot @WouterGlorieux for finding this!

I'll take care rebasing, pushing and releasing this ASAP.

rbval:
Is this fixed already in some version? Still experiencing this issue when using the `add_bytes` function.

alexander255:
Merged in #206. Sorry for the delay everybody!

alexander255:
This is now part of 0.4.13

rbval:
@alexander255 awesome thanks!

PR #203, opened 2020-04-04
Add develop install.

cottrell:
If this is not how you want to dev install, then let's add a note about how to do that in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,12 @@ Install with pip:
 pip install ipfshttpclient
 ```
 
+## Dev Install
+
+```sh
+python setup.py develop
+```
+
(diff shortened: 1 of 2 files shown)


alexander255:
@cottrell: Could you try running `flit install -e .` to perform the develop-install? I'm not well-versed with the specifics of this installation method, but I think this should achieve what you are trying to do without adding a dependency on `setup.py`/`setuptools`.

cottrell:
Cool! I had not heard to of that simple tool. Yes, it works like this.

    flit install --pth-file

alexander255:
Updated the README about this, thanks!

PR #205, opened 2020-04-13
typo in README.md regarding client.pin.ls example?

ingokeck:
The example states "client.pin_ls(...", but the correct form seems to be "client.pin.ls(..."


alexander255:
Thanks! `client.pin_ls` was the old *py-ipfs-api* syntax!

PR #207, opened 2020-04-17
Add username + password HTTP basic authentication support

### LINKED ISSUE #154: Basic auth support

alexander255:
Supersedes  and closes #155.

Fixes #154.

### DIFF
--- a/ipfshttpclient/http.py
+++ b/ipfshttpclient/http.py
@@ -24,6 +24,7 @@ import six
 
 from . import encoding
 from . import exceptions
+from . import utils
 PATCH_REQUESTS = (os.environ.get("PY_IPFS_HTTP_CLIENT_PATCH_REQUESTS", "yes").lower()
                   not in ("false", "no"))
 if PATCH_REQUESTS:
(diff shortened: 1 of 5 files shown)


ipfshttpclient/http.py
```diff
@@ -284,33 +285,34 @@ def _do_raise_for_status(self, response):
...
+	def _request(self, method, url, params, stream=False, files=None,
+	             headers={}, username=None, password=None, data=None,
+	             timeout=120):
 		if "close_conn_on_upload" in self.workarounds \
 		   and method.upper() not in ("GET", "HEAD"):  # pragma: no cover (workaround)
 			headers = headers.copy()
 			headers["Connection"] = "close"
 		
+		auth = None
+		print(username, password, username or password)
```
AliabbasMerchant, line 297: Possible remnant of local debugging?

PR #211, opened 2020-04-17
Add and fix Windows CI

alexander255:


alexander255:
This was a shitload of work – the only upshot being that while there were tons of issues, none of them were in the actual library, but only in the test and environment setup code.

PR #219, opened 2020-06-06
Add diff command to object

mike7c2:
Exposes the diff command from object: https://docs.ipfs.io/reference/http/api/#api-v0-object-diff


test/functional/test_object.py
```diff
@@ -101,4 +101,31 @@ def test_patch_set_data(client):
...
+	"""
+	result = client.object.diff(
+		"QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n",
+		"QmV4QR7MCBj5VTi6ddHmXPyjWGzbaKEtX2mx7axA5PA13G"
+	)
+	assert result == {'Changes': [{
+		'Type': 2,
+		'Path': '',
+		'Before': {'/': 'QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n'},
+		'After': {'/': 'QmV4QR7MCBj5VTi6ddHmXPyjWGzbaKEtX2mx7axA5PA13G'}}]}
```
ntninja, line 131: These tests depend on the contents of `fake_dir` having been added to the daemon first, right?

If so, please annotate them with a dependency on the test that adds this content to prevent spurious errors if this prerequirement isn't fulfilled or only a subset of tests has been run ([examples](https://github.com/ipfs-shipyard/py-ipfs-http-client/blob/5e3edf5f8e9e117738c77b5773c78d7b48942bc2/test/functional/test_files.py#L373-L422), [reference](https://pytest-dependency.readthedocs.io/en/stable/usage.html)):

```
@pytest.mark.dependency(depends=["test_add_recursive"])
```

ipfshttpclient/client/object.py
```diff
@@ -345,4 +345,36 @@ def stat(self, cid, **kwargs):
...
+			Key of object a for comparison
+		b : Union[str, cid.CIDv0, cid.CIDv1]
+			Key of object b for comparison
+
+		Returns
+		-------
+			dict
+		"""
+		args = (str(a), str(b))
+		return self._client.request('/object/diff', args, decoder='json', **kwargs)
```
ntninja, line 380: Thank you! :+1:

test/functional/test_object.py
mike7c2: Sorry for the delay on this, they do indeed :+1: 

I've added marks for the other tests that needed them too. 

I was in two minds as to whether to leave or remove the comments about the content dependency. I've left them in for now.

ntninja:
Thank you!

PR #223, opened 2020-06-27
Update versions for 0.6.0 release (synchronize with PyPI)

radfish:
Synchronize with the tarball that has been released on PyPI under a new name: ipfshttpclient4ipwb 0.6.0
I think what's missing is after merging this PR tag the master with 0.6.0, release it on github, and upload as a new release to pypi ipfshttpclient package (and perhaps delete ipfshttpclient4ipwb from PyPI to avoid confusion).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ dist: bionic  # VirtualEnv is too old on xenial
 
 matrix:
    include:
-      - python: "3.5"
-      - python: "3.6"
       - python: "3.7"
       - python: "3.8"
       - python: "pypy3"
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Check out [the HTTP Client reference](https://ipfs.io/ipns/12D3KooWEqnTdgqHnkkwa
 See the [relevant section of the README](#important-changes-from-ipfsapi-04x) for details.
 
 **Note:** This library occasionally has to change to stay compatible with the IPFS HTTP API.
-Currently, this library is tested against [go-ipfs v0.5.0-rc2](https://github.com/ipfs/go-ipfs/releases/tag/v0.5.0-rc2).
+Currently, this library is tested against [go-ipfs v0.6.0](https://github.com/ipfs/go-ipfs/releases/tag/v0.6.0).
 We strive to support the last 5 releases of go-IPFS at any given time; go-IPFS v0.4.21 therefore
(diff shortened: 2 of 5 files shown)


ntninja:
Thank you for opening this! Dropping support for 3.5 is not something I will consider until 2021-04 (when Ubuntu 16.04 is EOL). Keeping support for 3.5 is slightly annoying (variable typing annotations, incomplete async/await), but not overly problematic, so I see no reason to drop OOTB support early for those (few) users.

radfish:
That all sounds good. I am only showing the patch that was pushed to ipfshttpclient4ipwb on PyPI. That's the only tarball among releases on PyPI and Github that works with IPFS 0.6.0 daemon. Any idea on why that was pushed under a new name on PyPI and was not released on Github? I'm guessing you're not the owner of that new PyPI package then.. rogue release?

Please tread this PR as a question/suggestion for releasing the 0.6.0-compatible version on github and on PyPI under ipfshttpclient name, because the 0.4.13.2 does not work with 0.6.0. Thanks.

radfish:
Really confusing. So #222 suggests support for 0.6.0 is not finished, but yet there is this mystery release on PyPI labeled 0.6.0 (that is essentially current master except the patch in this PR) -- who owns this?:
https://pypi.org/project/ipfshttpclient4ipwb/

FWIW, I'm using essentially current master with 0.6.0: CI tests pass and all (old) functionality seems to be working. I guess #222 is about new functionality.

radfish:
```
2020-06-28 12:32:19     redfish ntninja: this is the mystery package i'm talking about: https://pypi.org/project/ipfshttpclient4ipwb/
2020-06-28 12:33:05     @ntninja        It's not that mysterious really. ;-)
2020-06-28 12:35:25     @ntninja        IPWB devs wanted me to release a new version with 0.5.x compatibility, but as is common knowledge now, I did not. However somewhere they wrote that one apparently cannot release packages on PyPI with git dependencies, so given the name they apparently decided to release a patched version for them to depend on on PyPI for the time being.
2020-06-28 12:40:23     redfish thanks. if they bothered to add this one sentence to the README on their PyPI fork, that would saved a lot of confusion.
2020-06-28 12:43:20     redfish anyway, with limited time to implement new functionality, what i would do is release current master under 0.4.13.3, so that people can use it with newest daemons, but without implying that all 0.6.0 functionality is available (that will come in 0.6.0 release of ipfshttpclient)
```

PR #224, opened 2020-06-27
test: print connection error

radfish:
To provide reason for failure when running tests.

Note: to see stderr/stdout from tests: ./run-tests.py --capture=no

### DIFF
--- a/test/functional/conftest.py
+++ b/test/functional/conftest.py
@@ -21,6 +21,7 @@ def is_available():  # noqa
 		try:
 			ipfshttpclient.connect()
 		except ipfshttpclient.exceptions.Error as error:
+			print("Failed to connect to IPFS daemon:", error)
 			__is_available = False
 			
 			# Make sure version incompatibility is displayed to users


ntninja:
What information does this provide that the test failure on `test_ipfs_node_available` does not provide? (If no daemon is available all functional tests will be skipped, but that one will fail with an error message.)

radfish:
Without this patch the failure output does not include the reason for the failure. For example, if test fails because daemon returned 500 response, you can't see that. There are many possible reasons for failure to connect to the daemon and it is useful to know which one happened, and very annoying to not have that information printed.

c0llab0rat0r:
Fixed on 4/17 in https://github.com/ipfs-shipyard/py-ipfs-http-client/commit/89737bde923f348aed1ab3a76a75c5c505bcc9d9

@ntninja this can be closed.

ntninja:
The 4/17 (in my mind: April 2017) confused me there… are you north american by any chance? :grinning: 

You're right on closing this through…

PR #238, opened 2020-09-07
Added optional arguments to `/dag/put`

meyer1994:
I am not sure if I did this correctly. I still do not understand some parts of the code. There are lots of backawards compatibility stuff that I do not know (yet).

### DIFF
--- a/ipfshttpclient/client/dag.py
+++ b/ipfshttpclient/client/dag.py
@@ -41,7 +41,7 @@ class Section(base.SectionBase):
 
 	@base.returns_single_item(base.ResponseBase)
 	def put(self, data: utils.clean_file_t, format: str = 'cbor',
-		    input_enc: str = 'json', **kwargs: base.CommonArgs):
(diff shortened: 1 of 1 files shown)


ipfshttpclient/client/dag.py
```diff
@@ -69,6 +70,8 @@ def put(self, data: utils.clean_file_t, **kwargs: base.CommonArgs):
 			dict
```
ntninja, line 70: The parameter section here should document the `format` and `input_enc` parameters

ntninja (approved):
Missing docs for the new parameters, otherwise LGTM!

(I'm currently a bit busy, so feel free to add the comments, otherwise I'll do when I merge this.)

ipfshttpclient/client/dag.py
meyer1994: Done :)

ntninja:
Thanks!

Issue #240, opened 2020-09-29
bug:  ipfshttpclient.exceptions.VersionMismatch: Unsupported daemon version '0.7.0' 


thelostone-mc:
#### Description 

[ipfshttpclient==0.6.0](https://github.com/gitcoinco/web/blob/master/requirements/ci.txt#L2)
[image: ipfs/go-ipfs:v0.6.0](https://github.com/gitcoinco/web/blob/master/docker-compose.yml#L45
)

CI link: https://travis-ci.org/github/gitcoinco/web/builds/731291982

PR where I'm trying to get it to pass by upgrading ipfshttpclient -> 0.6.1
https://github.com/gitcoinco/web/pull/7561

```
version = [0, 7, 0], minimum = [0, 4, 22], maximum = [0, 7, 0], blacklist = []
    def assert_version(version, minimum=VERSION_MINIMUM, maximum=VERSION_MAXIMUM, blacklist=VERSION_BLACKLIST):
    	"""Make sure that the given daemon version is supported by this client
    	version.
    
    	Raises
    	------
    	~ipfshttpclient.exceptions.VersionMismatch
    
    	Parameters
    	----------
    	version : str
    		The actual version of an IPFS daemon
    	minimum : str
    		The minimal IPFS daemon version to allowed
    	maximum : str
    		The maximum IPFS daemon version to allowed
    	"""
    	# Convert version strings to integer tuples
    	version = list(map(int, version.split('-', 1)[0].split('.')))
    	minimum = list(map(int, minimum.split('-', 1)[0].split('.')))
    	maximum = list(map(int, maximum.split('-', 1)[0].split('.')))
    
    	if minimum > version or version >= maximum:
>   		raise exceptions.VersionMismatch(version, minimum, maximum)
E     ipfshttpclient.exceptions.VersionMismatch: Unsupported daemon version '0.7.0' (not in range: 0.4.22 ≤ … < 0.7.0)
````

thelostone-mc:
Looks similar to https://github.com/ipfs-shipyard/py-ipfs-http-client/issues/239

ntninja:
It's fixed in the 0.7.0a1 prerelease.

Issue #239, opened 2020-09-29
ipfshttpclient.exceptions.VersionMismatch: Unsupported daemon version '0.7.0' (not in range: 0.4.22 ≤ … < 0.7.0)


hicetnunc2000:
I ran into this error from this code:
```
import ipfshttpclient
conn = ipfshttpclient.connect('/dns4/ipfs.infura.io/tcp/5001/https')
```
I'm using ```python 3.8.2, ipfshttpclient 0.6.1```.
even though it was working some days ago I'm wondering if there were any updates on go-ipfs/infura and if there is a way to adapt the package.

njgheorghita:
We're also blocked by this error in [web3.py](https://github.com/ethereum/web3.py/pull/1757). It appears to me that Infura recently updated their IPFS endpoint to version `7.0`, causing this error when you try to connect to their provider.

hicetnunc2000:
I've managed to connect with ipfs with an old ipfs-js-client though lol in case anyone is under some urgency

marcgarreau:
:wave: @ntninja @meyer1994 is #241 an acceptable solution?

ntninja:
Sorry for the delay! Version 0.7.0a1 fixes this for the time being until all the changes for the 0.7.0 version have completed. (The 0.7.0 release may include some backward-incompatible changes from 0.7.0a1).

PR #241, opened 2020-10-04
Test with go-ipfs v0.7.0 and drop v0.4.22

SupraSummus:
related #239 #240

My default editor config resulted in some whitespacing changed. I can undo that if it's unwelcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
       - python: "3.8"
         env: TOXENV=typeck
         before_install: ":"
-     
+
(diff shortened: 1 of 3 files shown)


ntninja:
Released a version of this without the whitespace changes as 0.7.0a1 – non-prerelease version will be released once I have all planned breaking changes for this release in. Sorry, for being slow in releasing compatibility bumps! The current versioning scheme makes this a bit annoying and I always hope to be done faster than I actually end up being.

PR #242, opened 2020-10-12
Update __init__.py

henrygong:
Since the current release version of the IPFS daemon is 0.7.0 and VERSION_MAXIMUM uses a < operator, there is a version mismatch error (see bugs #239 and #240 ). Simplest fix is bumping max to 0.7.1


ntninja:
Has been merged as part of #241. Thank you!

Issue #252, opened 2021-03-30
V0.8.0 support?


tanevanwifferen:
Can we just bump the required ipfs daemon version?

ntninja:
I'm aiming to commit a huge set of changes before releasing 0.8 (again, its been delayed several times already). Please ping me in a week if you haven't heard anything from me before that.

ntninja:
In general upgrading to a new go-IPFS works something like this (but do this yet):

  1. Update the minimum and maximum version in `ipfshttpclient/client/__init__.py` (next after 0.4.23 was 0.5.0)
  2. Update `.travis.yml` with matching go-ipfs versions for testing (both number and sha512 hash)
  3. Update `README.md` with matching version numbers in the note before the table of contents
  4. Open a PR
  5. Ensure that all checks (except for typing which is broken currently) pass, possibly requires patches and version workarounds to be written

^ Not all that bad, but it does take a bit of time.

dokterbob:
Hey all, I'd like to strongly suggest to change `VersionMismatch` from an Exception into a Warning, allowing users to use whatever version they'd like. I would consider the current behaviour 'broken by design' from a user perspective; most IPFS behaviour doesn't change between versions but the current behaviour ensures that it still breaks and that seems unnecessary to me.

Of course having full official 0.8.0 support and other changes are super welcome. It's just IPFS 0.8 has been out for a while now and this is a major pain for users, which will occur again and again.

ntninja:
I agree with this nowadays. Back in the 0.4.x days the breakage thing was pretty bad, but it's been a while and things seem to have settled down. Thank you for the suggestion!

tanevanwifferen:
@ntninja any update? we're doing manual backups since the latest ipfs release...

c0llab0rat0r:
@ntninja,

Also interested in a `0.8.0` release. Thanks.

0kalekale:
lads when are we getting v0.8.0 support

c0llab0rat0r:
@ntninja has put a lot of time and effort into this great library.

With the interest and activity on it, I wonder if it needs an additional owner/shepherd.

ntninja:
Somebody want to become co-maintainer? It appears I'm not as enthusiastic about maintaining this as I was when I took this over 5 years ago…

And 0.8.0 isn't the first release that has been greatly delayed for no strong reason either!

ntninja:
[...]

Issue #250, opened 2021-03-03
version mismatch


Narendhraprasth:
Unsupported daemon version '0.8.0' (not in range: 0.4.23 ≤ … < 0.8.0)

always some errors like this does happen can anyone help me ? i kind of tried the irc freenode but it doesn't work at all

marcusbeckerman:
I went into the package folder and opened `client/__init__.py` and changed the `VERSION_MAXIMUM` to  `"0.8.1"` on line 19, I haven't encountered any errors yet. Hopefully devs can push an update to officially support 0.8.0 soon.

c0llab0rat0r:
Version `0.8.0a1` has been published to PyPi.

kasteph:
This looks like a duplicate of #313 and it seems to be an ongoing problem whenever kubo upgrades. For any ongoing discussions/issues for this topic, I'd like to defer to the other issue so that it's easier to have the info in one place and focus.

Issue #234, opened 2020-09-03
version 0.6.1 failing with go-ipfs v0.6.0


octavioamu:
Im trying to run my tests and still getting 
`E     ipfshttpclient.exceptions.VersionMismatch: Unsupported daemon version '0.4.21' (not in range: 0.4.22 ≤ … < 0.7.0)`
ipfshttpclient 0.6.1
![image](https://user-images.githubusercontent.com/1358093/92178945-13626f80-ee1a-11ea-84f3-73dafef2e601.png)

go-ipfs 0.6.0
![image](https://user-images.githubusercontent.com/1358093/92178528-258fde00-ee19-11ea-8bcf-530e6afedf98.png)
Not sure If Im doing something wrong, or is a bug.

machawk1:
I am unable to replicate this with the same version of go-ipfs and ipfshttpclient. @octavioamu can you provide the code in your test that is raising this `VersionMismatch` exception in ipfshttpclient?

crzypatchwork:
same running 0.8.0

c0llab0rat0r:
Version `0.8.0a1` has been published to PyPi.

PR #253, opened 2021-03-31
Change exceptions.VersionMismatch into warning

dokterbob:
Let users decide whether they’re willing to take the risk; most of IPFS’ API doesn’t change between versions.

This will effectively allow users to use IPFS 0.8, at their own risk. Addresses #252, #250, #234
In addition, this will enable users to help with testing new versions.

I was unable to run tests locally and the tests in the default branch are broken. Manual evaluation of CI test results is required.

### DIFF
--- a/test/unit/test_client.py
+++ b/test/unit/test_client.py
@@ -1,23 +1,51 @@
+import warnings
+
 import pytest
 
 import ipfshttpclient
 
 
 def test_assert_version():
-	# Minimum required version
-	ipfshttpclient.assert_version("0.1.0", "0.1.0", "0.2.0", ["0.1.2"])
-	
+	with warnings.catch_warnings(record=True) as w:
+		# Cause all warnings to always be triggered.
+		warnings.simplefilter("always")
+
+		# Minimum required version
+		ipfshttpclient.assert_version("0.1.0", "0.1.0", "0.2.0", ["0.1.2"])
+
+		assert len(w) == 0
+
 	# Too high version
-	with pytest.raises(ipfshttpclient.exceptions.VersionMismatch):
+	with warnings.catch_warnings(record=True) as w:
+		# Cause all warnings to always be triggered.
+		warnings.simplefilter("always")
+
 		ipfshttpclient.assert_version("0.2.0", "0.1.0", "0.2.0", ["0.1.2"])
-	
+
+		assert len(w) == 1
+		assert issubclass(w[-1].category, ipfshttpclient.exceptions.VersionMismatch)
+
+
 	# Too low version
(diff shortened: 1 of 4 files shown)


dokterbob:
I've manually evaluated all the tests and anything working in `master` is working here as well, including the modified tests for `VersionMismatch` warnings. I think it's ready for merge (pending maintainer feedback).

Ref: https://travis-ci.com/github/ipfs-shipyard/py-ipfs-http-client/jobs/495154225#L375

ntninja:
I agree with this nowadays. Back in the 0.4.x days the breakage thing was pretty bad, but it's been a while and things seem to have settled down. Thank you for the suggestion!

The Windows error looks like they stopped distributing Python from Chocolatey. \*sigh\*
Will look more into this (and your PR) in the next couple of days.

0kalekale:
how long untill this gets upstreamed?

test/unit/test_client.py
```diff
@@ -1,24 +1,52 @@
...
 	# Blacklisted version
-	with pytest.raises(ipfshttpclient.exceptions.VersionMismatch):
+	with warnings.catch_warnings(record=True) as w:
+		# Cause all warnings to always be triggered.
+		warnings.simplefilter("always")
+
 		ipfshttpclient.assert_version("0.1.2-1", "0.1.0", "0.2.0", ["0.1.2"])
 
+		assert len(w) == 1
+		assert issubclass(w[-1].category, ipfshttpclient.exceptions.VersionMismatch)
```
ntninja, line 48: Please use [`pytest.warns`](https://docs.pytest.org/en/stable/reference.html#pytest-warns) as it makes for more easily readable code.

test/unit/test_client.py
```diff
@@ -1,24 +1,52 @@
...
-	ipfshttpclient.assert_version("0.1.0", "0.1.0", "0.2.0", ["0.1.2"])
-	
+	with warnings.catch_warnings(record=True) as w:
+		# Cause all warnings to always be triggered.
+		warnings.simplefilter("always")
+
+		# Minimum required version
+		ipfshttpclient.assert_version("0.1.0", "0.1.0", "0.2.0", ["0.1.2"])
+
+		assert len(w) == 0
```
ntninja, line 16: This first assertion can apparently also be shortend using `pytest.warns`: https://stackoverflow.com/a/45671804/277882

ntninja (changes requested):
Good! Tests can be shortened quite a bit using `pytest.warns` through.

c0llab0rat0r:
A rebase of this branch will fix its build failure.

Issue #255, opened 2021-03-31
Unbreak the mypy tests


dokterbob:
Making it easier to make (for contributors) and accept (for maintainers) contributions.

ntninja:
mypy stuff breaks with like every second update, but it should be fixed again I agree.

ntninja:
Want to delay this until after 0.8 since it's likely a lot of work.

c0llab0rat0r:
> Want to delay this until after 0.8 since it's likely a lot of work.

This will be significantly less work if the minimum Python version is raised to 3.7. https://github.com/ipfs-shipyard/py-ipfs-http-client/issues/258

c0llab0rat0r:
If #257 and #260 are merged, it would resolve this issue.

c0llab0rat0r:
@ntninja with the merge of #261, I believe this is now resolved, though there is more work to do to remove type hint warning suppression.

ntninja:
Yes it can be better or course, but it a much better situation than what we started out with thanks to you!

PR #257, opened 2021-04-01
Fix httpcore dependency; fix style linting

c0llab0rat0r:
1. Fix httpcore dependency by removing comment that broke pip
2. Add scripts to run tests and typing checks with Docker against multiple versions of Python

Previous to this change, attempts to run

    tox -e styleck -e typeck

..resulted in this error:

    ERROR: Invalid requirement: 'httpcore (~= 0.10.2)  # Has Unix domain socket support'

The error is caused by this part of that line being present in `tox.ini`:

    # Has Unix domain socket support

This commit does not fix the typing errors referenced in #255 but does demonstrate the typing errors now that the `httpcore` dependency is fixed.

### DIFF
--- a/ipfshttpclient/encoding.py
+++ b/ipfshttpclient/encoding.py
@@ -105,8 +105,9 @@ class Json(Encoding[utils.json_value_t]):
 		self._decoder2  = json.JSONDecoder()
 		self._lasterror = None  # type: ty.Optional[ValueError]
 	
-	@ty.no_type_check  # It works just fine and I don't want to rewrite it just
-	                   # because mypy doesn't understand…  # noqa: E114, E116
+	# It works just fine and I don't want to rewrite it just because mypy doesn't understand…
+	# noqa: E114, E116
+	@ty.no_type_check
 	def parse_partial(self, data: bytes) -> ty.Generator[utils.json_value_t, ty.Any, ty.Any]:
 		"""Incrementally decodes JSON data sets into Python objects.
 		
--- a/ipfshttpclient/encoding.py
+++ b/ipfshttpclient/encoding.py
@@ -106,7 +106,6 @@ class Json(Encoding[utils.json_value_t]):
 		self._lasterror = None  # type: ty.Optional[ValueError]
 	
 	# It works just fine and I don't want to rewrite it just because mypy doesn't understand…
-	# noqa: E114, E116
 	@ty.no_type_check
 	def parse_partial(self, data: bytes) -> ty.Generator[utils.json_value_t, ty.Any, ty.Any]:
 		"""Incrementally decodes JSON data sets into Python objects.
--- a/ipfshttpclient/filescanner.py
+++ b/ipfshttpclient/filescanner.py
@@ -512,7 +512,7 @@ class walk(ty.Generator[FSNodeEntry, ty.Any, None], ty.Generic[ty.AnyStr]):
 			#
 			# Note: `os.fwalk` support for binary paths was only added in 3.7+.
 			directory_str_or_fd = directory_str  # type: ty.Union[ty.AnyStr, int]
-			if HAVE_FWALK and (not isinstance(directory_str, bytes) or HAVE_FWALK_BYTES):  # noqa: E501
+			if HAVE_FWALK and (not isinstance(directory_str, bytes) or HAVE_FWALK_BYTES):
 				self._close_fd = directory_str_or_fd = os.open(directory_str, os.O_RDONLY | O_DIRECTORY)
 			
 			self._generator = self._walk(
--- a/ipfshttpclient/filescanner.py
+++ b/ipfshttpclient/filescanner.py
@@ -512,7 +512,7 @@ class walk(ty.Generator[FSNodeEntry, ty.Any, None], ty.Generic[ty.AnyStr]):
 			#
 			# Note: `os.fwalk` support for binary paths was only added in 3.7+.
 			directory_str_or_fd = directory_str  # type: ty.Union[ty.AnyStr, int]
-			if HAVE_FWALK and (not isinstance(directory_str, bytes) or HAVE_FWALK_BYTES):  # type: ignore[unreachable]  # noqa: E501
+			if HAVE_FWALK and (not isinstance(directory_str, bytes) or HAVE_FWALK_BYTES):  # noqa: E501
 				self._close_fd = directory_str_or_fd = os.open(directory_str, os.O_RDONLY | O_DIRECTORY)
 			
 			self._generator = self._walk(
@@ -583,7 +583,7 @@ class walk(ty.Generator[FSNodeEntry, ty.Any, None], ty.Generic[ty.AnyStr]):
 	) -> ty.Generator[FSNodeEntry, ty.Any, None]:
 		if directory_str is not None:
 			sep = utils.maybe_fsencode(os.path.sep, directory_str)
-		elif matcher is not None and matcher.is_binary:  # type: ignore[unreachable]
+		elif matcher is not None and matcher.is_binary:
 			sep = os.fsencode(os.path.sep)  # type: ignore[assignment]
 		else:
 			sep = os.path.sep  # type: ignore[assignment]
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -173,15 +173,21 @@ try:
 				exclusions.append(r"\# pragma: http-backend=httpx")
 			
 			# Create temporary file with extended *coverage.py* configuration data
-			coveragerc.file.writelines(map(lambda s: s + "\n", itertools.chain((
-				"[run]",
-				"omit =",
-			), map(lambda s: "\t" + s, omitted_files),
-				(
-					"[report]",
-					"# Exclude lines specific to some other Python version from coverage",
-					"exclude_lines =",
-				), map(lambda s: "\t" + s, exclusions))))
+			coveragerc.file.writelines(
+				map(
+					lambda s: s + "\n",
+					itertools.chain(
+						(
+							"[run]",
+							"omit =",
+						),
+						map(lambda s: "\t" + s, omitted_files),
+						(
+							"[report]",
+							"# Exclude lines specific to some other Python version from coverage",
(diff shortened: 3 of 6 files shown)


ntninja:
Hmm… I never considered the typing and style checks to be particularily hard to run. Using docker to run them seems kinda overkill to me, but if people feel this is helpful I'll add this.

However, before we do that, could you tell me what output you are seeing when running the `tox -e styleck -e typeck` command directly on your system? Also what does `tox --version` report? (Tox has a feature where it'll provision a newer version of itself, if the host version is too old, so I'd strongly prefer using that instead if that is all there is to it.)

c0llab0rat0r:
I didn't have tox installed at all. In the python:3.8.5 docker image its 3.7.0, and when I install tox into my venv its 3.23.0.

The tox output, both from the docker runs and with tox now installed into the venv, is two categories of errors.

Linting, such as:

    ./venv/lib/python3.8/site-packages/netaddr/strategy/eui48.py:267:1: WT191 (flake8-tabs) indentation contains spaces, but option indent-style=tab requires tabs

And typing, such as:

ipfshttpclient/client/base.py:212: error: Incompatible return value type (got
"Callable[[VarArg(Any), KwArg(Any)], Union[None, List[R], ResponseWrapIterator[T, R]]]", expected "_returns_multiple_wrapper2_t[T, R]") 
[return-value]
                    return wrapper2
                           ^
i

c0llab0rat0r:
The primary benefit of the docker images is validating against the array of supported python versions easily with a clean build. You could check them individually with separate virtual environments, but then you have to maintain the virtual environments and dependencies could drift. Docker is a clean slate each time.

c0llab0rat0r:
I think there's a combination of a small number of legitimate issues the type and style checks are catching, and a larger number of false positives caused by a lack of ignores for checking dependencies (e.g. warnings for `venv/../code.py`).

I'm going to make some adjustments and push a revision.

ntninja:
Thank you, now I understand much better what this issue is about! I thought this was merely to workaround some issue you had with `tox`, but using this for cross-version checks makes much more sense. (I have a setup for this without using docker, but that is not something most people can be asked to reproduce.)

verify.sh
```diff
@@ -0,0 +1,27 @@
...
+    docker run -it -v $(pwd):/source:ro -w /usr/src/app $tag tox -e styleck -e typeck
+}
+
+if [ -z "$1" ]; then
+    echo "Validating supported versions..."
+
+    validate 3.5.10
+    validate 3.6.13
+    validate 3.7.10
+    validate 3.8.8
```
ntninja: Is is possible to use aliases like `3.5`, `3.6`, … here to avoid listing all these detailed versions?

verify.sh
```diff
@@ -0,0 +1,27 @@
...
+else
+    echo "Validating only $1..."
```
ntninja: Nit: Please use ellipsis (…), instead of three dots (...) here.

verify.sh
```diff
@@ -0,0 +1,27 @@
...
+
+function validate() {
+    python_version=$1
+
+    echo "Validating version $python_version..."
+
+    tag=py-ipfs-http-client-verify:$python_version
+
+    docker build --build-arg PYTHON_VERSION=$python_version -t $tag .
+    docker run -it -v $(pwd):/source:ro -w /usr/src/app $tag tox -e styleck -e typeck
```
ntninja: Use `docker run --rm` to ensure the used image doesn't stick around needlessly.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+ARG PYTHON_VERSION
+
+FROM python:${PYTHON_VERSION}
+
+RUN apt-get update -y && apt-get install -y apt-utils tox
```
ntninja: Would it be OK to use the Python `-alpine` images here for smaller image download size?

Install command would be `apk update && apk add --no-cache py3-tox` in this case.

verify.sh
```diff
@@ -0,0 +1,27 @@
...
+}
+
+if [ -z "$1" ]; then
+    echo "Validating supported versions..."
```
ntninja: Nit: Please use ellipsis (…), instead of three dots (...) here.

entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+cp -r /source/* /usr/src/app/
+
+exec $@
+
```
ntninja: Can this script be inlined into the `Dockerfile` (less clutter)? Also `#!/bin/sh` is sufficient here.

verify.sh
ntninja: Please always add double-quotes (") around variable/process substitutions (ie: `$tag` → `"${tag}"`) unless you explicitly intend to invoke shell word splitting. Figuring out whitespace issues with shell is a huge PITA otherwise.

test/run-tests.py
```diff
@@ -177,11 +177,11 @@ def pytest_pyfunc_call(self, pyfuncitem):
 				"[run]",
 				"omit =",
 			), map(lambda s: "\t" + s, omitted_files),
-			(
-				"[report]",
-				"# Exclude lines specific to some other Python version from coverage",
-				"exclude_lines =",
-			), map(lambda s: "\t" + s, exclusions))))
+				(
```
ntninja, line 180: It will behave better here if you move the opening parenthese to the line above. The way it's currently written was actually a workaround to a bug in an older version of the checker. :wink:

ntninja (commented):
You just dropped all the Docker-related code I tried to review. What is the plan there?

verify.sh
c0llab0rat0r: You can; it's a trade off between build consistency and being automatically on a later point release of Python.

Which would you prefer?

I think ideally it would use the same source of truth as the rest of the build and scrape the config files, but I didn't get that far.

ntninja:
As a more general note on the Docker thing: It is not really necessary to run the type and style checks in different Python versions. CI only checks the latest version and this is on purpose.

For typing, lot's of workarounds were required for pre-3.8 versions type hints, so the hints are provided on a best-effort basis there. The reason for running the type check at all is to ensure our typings are non-contradictionary and this property should not change depending on the 3.8+ Python version run (many mypy bugs permitting), so checking more than one Python version isn't all that helpful.

For flake8 the Python version is pretty much totally irrelevant, except that the version needs to be new enough to understand all the syntax used in the project's source code (`ast` and `tokenize` modules). Again, this is the case for any Python version since 3.6, so no need to run the check on more than one version.

verify.sh
c0llab0rat0r: > You just dropped all the Docker-related code I tried to review. What is the plan there?

I thought I was getting some push back on it so I decided to pare down the PR to changes to the library. I can add it back if you like, with changes per the comments you added.

FYI - the linting step of the Travis build now passes. I don't have fixes for the remaining type checks since I'm not familiar enough with protocols to address it.

Do you want me to add the Docker scripts back into this PR, or open a new PR for it?

c0llab0rat0r:
Linting step for 3.8 build is passing now.

verify.sh
ntninja: Consistency is probably better, you're right! It's probably best to test the supported *earliest* version of each Python minor release.

> I think ideally it would use the same source of truth as the rest of the build and scrape the config files, but I didn't get that far.

The current “single source of truth” would be the `tool.flit.metadata.requires-python` key in `pyproject.toml`, whose value would effectively evaluate to the following Python minimum minor version configuration:

```sh
    validate 3.5.4
    validate 3.6.2
    validate 3.7.2
    validate 3.8.0
```

It's not straightforward at all to convert these values programmatically through. :-/

verify.sh
ntninja: Please open a new PR for the Docker stuff. The current PR is indeed useful as is, so no need to clutter it further.

ipfshttpclient/encoding.py
```diff
@@ -105,8 +105,9 @@ def __init__(self) -> None:
 		self._decoder2  = json.JSONDecoder()
 		self._lasterror = None  # type: ty.Optional[ValueError]
 	
-	@ty.no_type_check  # It works just fine and I don't want to rewrite it just
-	                   # because mypy doesn't understand…  # noqa: E114, E116
+	# It works just fine and I don't want to rewrite it just because mypy doesn't understand…
+	# noqa: E114, E116
```
ntninja, line 109: Can you check if these `noqa` items are still needed after reformatting?

ipfshttpclient/filescanner.py
```diff
@@ -512,7 +512,7 @@ def __init__(
 			#
 			# Note: `os.fwalk` support for binary paths was only added in 3.7+.
 			directory_str_or_fd = directory_str  # type: ty.Union[ty.AnyStr, int]
-			if HAVE_FWALK and (not isinstance(directory_str, bytes) or HAVE_FWALK_BYTES):  # type: ignore[unreachable]  # noqa: E501
+			if HAVE_FWALK and (not isinstance(directory_str, bytes) or HAVE_FWALK_BYTES):  # noqa: E501
```
ntninja, line 515: Same here: `noqa: E501` is “line too long” and was added mostly likely because the mypy `type: ignore` comment blew up the line length in the first place.

test/run-tests.py
ntninja: ^ Do this.

ntninja (changes requested):
Just some minor stuff now.

ipfshttpclient/filescanner.py
c0llab0rat0r: Removed.

ipfshttpclient/encoding.py
c0llab0rat0r: Removed `noqa`.

test/run-tests.py
c0llab0rat0r: I had tried simply moving the parenthesis on line 180 up earlier but it didn't like that. The latest change flattens that entire section and passes the linter.

ntninja (approved):
Thank you!

Issue #258, opened 2021-04-04
Python 3.5 is past EOL; Drop Support


c0llab0rat0r:
Drop support for Python 3.5 by raising the minimum version to 3.6 (EOL this year) or 3.7 (EOL 2023).

This will make fixing the typing and type linting easier.

The code base currently has syntax that is not valid in 3.5, such as this at https://github.com/ipfs-shipyard/py-ipfs-http-client/blob/0.7.0/ipfshttpclient/http_httpx.py#L41

    def map_args_to_httpx(
        *,
        auth: auth_t = None,
        cookies: cookies_t = None,
        headers: headers_t = None,
        params: params_t = None,
        timeout: timeout_t = None,
    ) -> RequestArgs:

c0llab0rat0r:
If #260 is merged, it would resolve this issue.

ntninja:
The post-EOL compatibility with Python 3.5 is intentional since this project pledged support for Ubuntu 16.04LTS for servers. That will be EOL the end of this month, so dropping 3.5 is OK. Support for 3.6 is still necessary – in general this library does intend to be appealing to a large audience, so premature deprecation is a no-go.

The quoted piece of source code is not used in Python 3.5 btw, so it's lot like “Python 3.5 is already broken anyways” (this part of CI is still functional).

ntninja:
This you have fixed as well now!

PR #259, opened 2021-04-04
Add scripts to verify compatibility with multiple Python versions

c0llab0rat0r:
Carved out of https://github.com/ipfs-shipyard/py-ipfs-http-client/pull/257

Adds scripts that use Docker to verify the library's compatibility with the versions of Python it supports.

With this change, the `travis` build will still fail with style linting and type linting errors. Change #257 will correct the style linting errors, and this change will be rebased once that one is merged.

Once #257 and this change are both merged, the scripts in this change will fail on the first Python version, `3.5.4`, because the current code base has syntax at https://github.com/ipfs-shipyard/py-ipfs-http-client/blob/52c1d2a28a43fb020d4d6375be2b8d432267c37d/ipfshttpclient/http_httpx.py#L48 that is not supported by `3.5.4` - this problem will be corrected by #258.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,8 @@ description-file = "README.md"
 #        `typing.NoReturn` function signature support. (Also, many other `typing` module
 #        items were only introduced post-release in 3.6 and version restrictions on these
 #	     versions ensure that those are all available as well.)
+#
+#        Maintain this concurrently with verify.sh
 requires-python = ">=3.6.2,!=3.7.0,!=3.7.1"
 requires = [
 	"multiaddr (>=0.0.7)",
--- a/tools/verify/Dockerfile
+++ b/tools/verify/Dockerfile
@@ -1,0 +1,15 @@
+ARG PYTHON_VERSION
+
+FROM python:${PYTHON_VERSION}
+
+RUN pip install --upgrade pip
+RUN pip install tox
+
+# Mount the source code here, instead of to /usr/src/app.
+# Otherwise, tox will fail due to folder being read-only.
+# Mount only the source code; avoid mounting working folders.
+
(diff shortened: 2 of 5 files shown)


c0llab0rat0r:
Build error for `Windows` appears transient:

    # Daemon should not be running anymore
    >   	assert not daemon_is_running()
    E    assert not True
    E     +  where True = <function test_daemon_stop.<locals>.daemon_is_running at 0x000001BD05C99820>()

c0llab0rat0r:
`./verify.sh 3.8.0` will pass.

`./verify.sh 3.7.2` and `3.6.2` will fail, with problems related to `flake8` and `typing.Final` not being present in versions of Python prior to 3.8.

c0llab0rat0r:
Unrelated transient failure on existing build process on Windows; does not exercise new scripts which are only run manually and locally.

c0llab0rat0r:
Added a commit to add support for Python 3.9 since that was added to the library dependencies and build recently.

Using these new scripts:
* Python 3.6.2 will fail - flake fails to load a flake plugin that uses typing.Final, which is not supported on Python 3.6
* Python 3.7.2 will fail - this library uses Literal which is not supported on Python 3.7
* Python 3.8.0 will pass
* Python 3.9.0 will pass

We might be using a version of `flake` that is too new to use with Python 3.6.

c0llab0rat0r:
@ntninja 

Transient windows failure

https://travis-ci.com/github/ipfs-shipyard/py-ipfs-http-client/jobs/498603490

c0llab0rat0r:
Rebased from master; ready to go.

ntninja:
Thanks! :slightly_smiling_face:

PR #262, opened 2021-04-08
Bump min Python version from 3.5.4 to 3.6.2

c0llab0rat0r:
Reduction in scope of #260, since this client is pledged to support Python 3.6 thru its EOL.

#261 should be merged before this PR is.

PR #264, opened 2021-04-09
Miscellaneous cleanup from #260 and #262

c0llab0rat0r:


c0llab0rat0r:
@ntninja 

tiny change; is ready for review. 

The Windows build failed on several PRs with transient docker errors. This one is finally green again.

ipfshttpclient/client/__init__.py
```diff
@@ -106,6 +107,10 @@ def connect(
 	All parameters are identical to those passed to the constructor of the
 	:class:`~ipfshttpclient.Client` class.
 	"""
+
+	if headers is None:
+		headers = {}
+
```
ntninja, line 113: Quick question: What is the benefit of splitting this out over keeping it in inline in the declaration?

ntninja:
> The Windows build failed on several PRs with transient docker errors. This one is finally green again.

You have any logs of that happening? I'd like to investigate this if its a recurring problem.

ipfshttpclient/client/__init__.py
c0llab0rat0r: Mutable default values are dangerous because if the object gets mutated, the default value is different on the next call.

https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument
https://towardsdatascience.com/python-pitfall-mutable-default-arguments-9385e8265422

c0llab0rat0r:
> > The Windows build failed on several PRs with transient docker errors. This one is finally green again.
> 
> You have any logs of that happening? I'd like to investigate this if its a recurring problem.

It was likely caused by the build environment, and seemed to go away after about 30 minutes.

Here's a direct link: https://github.com/ipfs-shipyard/py-ipfs-http-client/runs/2309483883

If you browse up to the commit history directly above, you'll see a red `X` you can click on to show a link to the failing build.

ipfshttpclient/client/__init__.py
ntninja: I'd like to argue that its not a problem here since it will *never* mutate the given `headers` object without copying it. That could break things like:

```py
headers = {…}
client.add("abc", headers=headers)
client.add("xyz", headers=headers)
```

What should definitely change here is the definition of `headers_t`, which currently refers to `ty.Dict[…]` but should actually be `ty.Mapping[…]`. The `{}` here just represents an arbitrary empty mapping; if [`frozenmap`](https://www.python.org/dev/peps/pep-0603/) were a thing, I'd have used that instead.

If the code every mutates `headers` it is already wrong for the above reason.

ntninja:
> > > The Windows build failed on several PRs with transient docker errors. This one is finally green again.
> > 
> > 
> > You have any logs of that happening? I'd like to investigate this if its a recurring problem.
> 
> It was likely caused by the build environment, and seemed to go away after about 30 minutes.
> 
> Here's a direct link: https://github.com/ipfs-shipyard/py-ipfs-http-client/runs/2309483883

Oh **that** error! The CI environment was looking for new definitions of quite slow, I see.
This should hopefully “fix” it: 3d399d33302c004122773af52a740da1c736540e

> If you browse up to the commit history directly above, you'll see a red `X` you can click on to show a link to the failing build.

That's the thing, all the commits I see here (from the force-pushes and such) are all green, so thanks for the link.

c0llab0rat0r:
Uhgh, Windows and docker transient failures again.

@ntninja  here is the link https://travis-ci.com/github/ipfs-shipyard/py-ipfs-http-client/jobs/497516138

c0llab0rat0r:
@ntninja another transient Windows/docker build failure.

This change is otherwise ready for re-review.

https://travis-ci.com/github/ipfs-shipyard/py-ipfs-http-client/jobs/497523061

c0llab0rat0r:
@ntninja build is green again.

c0llab0rat0r:
Rebased with `master` - no other changes.

c0llab0rat0r:
Ready to merge

PR #266, opened 2021-04-09
Target Python 3.9 and go-IPFS 0.8.0

ntninja:


c0llab0rat0r:
Oh great a new Python version to support with more edges :) bring it on :)

ipfshttpclient/client/__init__.py
```diff
@@ -14,9 +14,9 @@
 DEFAULT_ADDR = multiaddr.Multiaddr(os.environ.get("PY_IPFS_HTTP_CLIENT_DEFAULT_ADDR", '/dns/localhost/tcp/5001/http'))
 DEFAULT_BASE = str(os.environ.get("PY_IPFS_HTTP_CLIENT_DEFAULT_BASE", 'api/v0'))
 
-VERSION_MINIMUM   = "0.4.23"
+VERSION_MINIMUM   = "0.5.0"
 VERSION_BLACKLIST = []
-VERSION_MAXIMUM   = "0.8.0"
+VERSION_MAXIMUM   = "0.9.0"
```
c0llab0rat0r, line 19: Why is this `0.9.0` and not `0.8.0` ?

Don't we want the warning behavior when connecting to `0.9.0`?

ipfshttpclient/client/base.py
```diff
@@ -167,25 +166,18 @@ def close(self) -> None:
...
-	    -> ty.Union[ty.List[T], http.StreamDecodeIteratorSync[T], None]:
+	    -> ty.Union[ty.List[T], http.StreamDecodeIteratorSync[T]]:
 		...
 
 
 class _returns_multiple_wrapper2_t(ty_ext.Protocol, ty.Generic[T, R]):
 	@ty.overload
-	def __call__(self, *args: ty.Any, stream: bool = ...,
-	             return_result: utils.Literal_False, **kwargs: ty.Any) -> None:
+	def __call__(self, *args: ty.Any, stream: utils.Literal_True, **kwargs: ty.Any) -> ty.List[R]:
```
c0llab0rat0r, line 175: I'd like to propose that `Literal[True]` and `Literal[False]` not be used until the minimum Python version is `3.8`

There's quite a few typing edge cases to deal with already related to the documentation tools having gaps, and supporting multiple Python versions. This adds overhead for other contributors and maintainers.

What benefit is being gained by using `Literal[t/f]` instead of `bool` ?

c0llab0rat0r (commented):
Should the build have automated tests that run against each supported "major" version of the IPFS server?

In this case, 0.5.x, 0.6.x, 0.7.x, and 0.8.x?

ipfshttpclient/client/__init__.py
ntninja: The check is `VERSION_MINIMUM <= version < VERSION_MAXIMUM` since it couldn't otherwise cover the entire version range of the maximum supported version series. (Think of it like Python's `range` function. :wink:)

ipfshttpclient/client/base.py
ntninja: > I'd like to propose that Literal[True] and Literal[False] not be used until the minimum Python version is 3.8

Do remember that this means end of 2023 for this library at the earliest.

> What benefit is being gained by using Literal[t/f] instead of bool?

In cases like this one, it allows mypy to distinguish between the two boolean values and have conflict-free overloads (otherwise the overload signatures would be identical). Also this overloading thing will have to be present in the public API as well to provide proper return types for methods actually supporting `stream` (only a couple fortunately) and the distinction will again be just the boolean value.

ntninja:
> Should the build have automated tests that run against each supported "major" version of the IPFS server?
> 
> In this case, 0.5.x, 0.6.x, 0.7.x, and 0.8.x?

I wanted to avoid adding even more test variations. With this, it tests against 0.5.0 and 0.8.0 on Linux and only 0.8.0 on macOS/Windows.

Travis isn't always as snappy as its been today/yesterday.

ipfshttpclient/client/__init__.py
c0llab0rat0r: I'd add a comment about it being an inclusive lower bound and exclusive upper bound.

ipfshttpclient/client/base.py
c0llab0rat0r: Knowing the EOL was 2023 was actually a part of what motivated my comment - another view is that it will be a drag factor on the library for 3 years.

Can the method be refactored another way, for example, having two methods (different names)? If the return type is correlated 1:1 with a true/false boolean, that seems to be a reason to have two separate methods/functions.

c0llab0rat0r:
> > Should the build have automated tests that run against each supported "major" version of the IPFS server?
> > In this case, 0.5.x, 0.6.x, 0.7.x, and 0.8.x?
> 
> I wanted to avoid adding even more test variations. With this, it tests against 0.5.0 and 0.8.0 on Linux and only 0.8.0 on macOS/Windows.
> 
> In my experience, Travis isn't always as snappy as it has been today/yesterday.

After I made my comment I wondered if the version requirement should be inverted.

Can we expect the IPFS server to support older clients? E.g. can we expect a 0.9 server to support 0.7 and 0.8 clients? This is the typical behavior of a stable api - support lower versioned clients for a limited period of time so they have time to upgrade.

Put another way, should the version check for the 0.8 pyipfshttpclient require a server version no less than 0.8, and prohibit a 0.7 server?

ntninja:
> Can we expect the IPFS server to support older clients? E.g. can we expect a 0.9 server to support 0.7 and 0.8 clients?

No, sadly the API is not stable. Although things have gotten a lot better since go-IPFS 0.5.0 was released, backwards compatibility between minor versions is still best-effort. I already agreed to change the error to a warning, since things are much better than they used to be and using a newer daemon version is pretty unlikely to cause issues at this point.

c0llab0rat0r:
> > Can we expect the IPFS server to support older clients? E.g. can we expect a 0.9 server to support 0.7 and 0.8 clients?
> 
> No, sadly the API is not stable. 

Ok, that makes sense.

It is after all a `0.x` :smile:

ipfshttpclient/client/base.py
ntninja: I think it's fine to have this for a dynamic language like Python and it's been like this for several years already. I think it's not worth redesigning the API just because it makes the typings a bit more verbose (and a bit ugly) in some places.

PR #268, opened 2021-04-10
Print exception thrown while trying to connect to IPFS 

c0llab0rat0r:
Inspired by #224 and the aggravation of the recent transient build failures connecting to the docker daemon, though that is a separate issue. It felt the same :@

* Print exception (on a double new line due to progress percentages)
* Refactor to use pytest fixtures instead of an evil evil `global` variable
* Remove some comments that restate `pytest` documentation about `fixture` scopes

### DIFF
--- a/test/functional/conftest.py
+++ b/test/functional/conftest.py
@@ -1,8 +1,10 @@
 # Note that this file is special in that py.test will automatically import this file and gather
 # its list of fixtures even if it is not directly imported into the corresponding test case.
-import pathlib
 
+import pathlib
 import pytest
+import sys
+import typing as ty
 
 import ipfshttpclient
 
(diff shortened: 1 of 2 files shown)


test/functional/conftest.py
```diff
@@ -1,78 +1,68 @@
...
-		yield client
+	yield from _generate_client(ipfs_is_available, False)
 
 
 @pytest.fixture(scope="module")
 def module_offline_client():
-	"""Create a client in offline mode with module lifetime."""
-	with get_client(offline=True) as client:
-		yield client
+	yield from _generate_client(ipfs_is_available, True)
```
ntninja, line 65: How do these work given that they don't actually depend on the `ipfs_is_available` fixture?

ntninja (approved):
Looks good!

Please just explain to dumb me how this fixture business works with alternate client fixture. :stuck_out_tongue_closed_eyes:

test/functional/conftest.py
c0llab0rat0r: Long form: https://docs.pytest.org/en/stable/fixture.html

tldr:

The style used for the IPFS client is a generator style. For example, for a fixture with `function` scope (`scope='function' - the default), when a test runs, pytest will start the fixture generator first, then at the point the generator `yield`s, pytest will call the test function, passing the fixture as an argument. When the test function completes, control is returned back to the fixture at the yield statement where it continues where it left off.

This style is useful for fixtures that provide values requiring cleanup (database connections, file handles, IPFS server connections, etc.).

From a test function, the fixture value is accessed by declaring an argument with the same name as the fixture function. For example:

    @pytest.fixture(scope='module')
    def foo():
        return 'some value used by the tests'

    def test_bar(foo):
        assert foo = 'some value used by the tests'

ntninja: Thanks, I knew that, but how does a fixture like this:

```
@pytest.fixture(scope="module")
def module_offline_client():
	"""Create a client in offline mode with module lifetime."""
	yield from _generate_client(ipfs_is_available, True)
```

… work if it *uses* the name `ipfs_is_available`, but doesn't actually take `ipfs_is_available` an argument/fixture-name.

Thinking about it again, I think it uses fixture definition `ipfs_is_available` from global scope, which, as a function, is always truethy. So it will unconditionally connect to the daemon, even when it is not actually available.

I think this is a bug and not some pytest magic I just didn't understand like I first assumed. :grin:

c0llab0rat0r: In the `foo/bar` example above, the `foo` argument of `test_bar` is a `bool` not a `Callable` - `pytest` evaluates the fixture (in this case, by simply calling it, because `foo` is not a generator, but a plain `Callable`), and then pytest will pass the return value to whatever tests are in scope (in this case, any test in the same module as the first test in a module that used the fixture, due to `scope='module'`).

ntninja: Yes, but…

test/functional/conftest.py
```diff
@@ -1,78 +1,68 @@
...
-	automatically using a fixture (to keep client creation code centralized
-	here), that client-creating fixture must also be module-scope, so use
-	this fixture in module-scoped fixtures."""
-	with get_client() as client:
-		yield client
+	yield from _generate_client(ipfs_is_available, False)
 
 
 @pytest.fixture(scope="module")
 def module_offline_client():
```
ntninja, line 64: … aren't you missing a parameter named `ipfs_is_available` on this line to have pytest resolve the fixture?

test/functional/conftest.py
c0llab0rat0r: I love the dramatic pause. It made me feel really good about myself when I saw the problem.

Thanks for that :smile:

test/functional/conftest.py
c0llab0rat0r: Yes, I am :sob:

c0llab0rat0r:
Ready to go.

This change should be squashed on merge.

PR #270, opened 2021-04-10
Add configuration scripts for local IPFS server

c0llab0rat0r:
* Add scripts to enable experimental features required by functional tests
* Adjust README


ntninja:
I think is may be a misunderstanding here: `tox -e py3` will start `test/run-tests.py` which already starts and configures the IPFS daemon (if in `${PATH}`), and also sets up coverage reporting since that is quite annoying to do, before invoking `pytest` in process with a special plugin for the `test_shutdown` test.

c0llab0rat0r:
> `tox -e py3` will start `test/run-tests.py` which already starts and configures the IPFS daemon `

I didn't see that behavior, I'll take a look again.

c0llab0rat0r:
> > `tox -e py3` will start `test/run-tests.py` which already starts and configures the IPFS daemon `
> 
> I didn't see that behavior, I'll take a look again.

Ok, so the misunderstanding here is that I was running the entire test suite from PyCharm, which doesn't run the shell script referenced by `tox` - it expects `pytest` to take care of all fixture and dependency setup and tear down.

https://pypi.org/project/pytest-docker/ is a more generalized solution to this problem that makes test execution more flexible.

PR #274, opened 2021-04-24
Fixed: assert_version was thinking that current version is not supported

miratcan:


c0llab0rat0r:
@miratcan your change has a failing unit test.

miratcan:
> @miratcan your change has a failing unit test.

I will fix asap

ntninja:
Thanks for trying to help, but there is nothing wrong here: The version on master should accept go-IPFS 0.8.x just fine and it not tested with 0.9.x yet. This is probably just another misunderstanding since I *still* didn't release it yet …

Issue #275, opened 2021-04-30
VersionMismatch: Unsupported daemon version '0.8.0' (not in range: 0.4.23 ≤ … < 0.8.0)


iskanderandrews:
Hi, 
I'm trying to connect with local ipfs daemon, but getting this error. 
`ipfshttpclient.exceptions.VersionMismatch: Unsupported daemon version '0.8.0' (not in range: 0.4.23 ≤ … < 0.8.0)`

```bash 
$ ipfs daemon 
```

```
Initializing daemon...
go-ipfs version: 0.8.0-ce693d7e8
Repo version: 11
System version: amd64/linux
Golang version: go1.16.2
2021/04/30 07:30:58 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
Swarm listening on /ip4/127.0.0.1/tcp/4001
Swarm listening on /ip4/127.0.0.1/udp/4001/quic
Swarm listening on /ip4/172.17.0.1/tcp/4001
Swarm listening on /ip4/172.17.0.1/udp/4001/quic
Swarm listening on /ip4/192.168.1.104/tcp/4001
Swarm listening on /ip4/192.168.1.104/udp/4001/quic
Swarm listening on /ip6/::1/tcp/4001
Swarm listening on /ip6/::1/udp/4001/quic
Swarm listening on /ip6/fd74:ae1:c698:db00:da5c:3a2c:f128:5c03/tcp/4001
Swarm listening on /ip6/fd74:ae1:c698:db00:da5c:3a2c:f128:5c03/udp/4001/quic
Swarm listening on /p2p-circuit
Swarm announcing /ip4/127.0.0.1/tcp/4001
Swarm announcing /ip4/127.0.0.1/udp/4001/quic
Swarm announcing /ip4/192.168.1.104/tcp/4001
Swarm announcing /ip4/192.168.1.104/udp/4001/quic
Swarm announcing /ip6/::1/tcp/4001
Swarm announcing /ip6/::1/udp/4001/quic
API server listening on /ip4/127.0.0.1/tcp/5001
WebUI: http://127.0.0.1:5001/webui
Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/8081
Daemon is ready

```

```bash 
$ pip show ipfshttpclient  
```

```
[...]

PR #276, opened 2021-05-01
Not accurate note statement in README

iskanderandrews:
Hi, 
According to the second Note in the README documentation: 

> Note: This library occasionally has to change to stay compatible with the IPFS HTTP API. Currently, this library is tested against _go-ipfs v0.8.0_. We strive to support the last 5 releases of go-IPFS at any given time; go-IPFS v0.5.0 therefore being to oldest supported version at this time.

So from this line: 
> Currently, this library is tested against _go-ipfs v0.8.0_. 

I understood that's working successfully with _go-ipfs v0.8.0_, and when tried to connect found this issue here #275  and it's still doesn't support  _go-ipfs v0.8.0_ until resolving this issue #274 , also this issue has been first mentioned in issue #252 

So I've changed in this PR the above note statement to be more accurate from my point of view: 
> Note: This library occasionally has to change to stay compatible with the IPFS HTTP API. Currently, this library is still under testing against go-ipfs v0.8.0. We strive to support the last 5 releases oftested go-IPFS at any given time; go-IPFS v0.5.0 therefore being to oldest supported version at this time.

Maybe I misunderstood it from the first time but actually didn't find meaning other than it's working with v8.
Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ unmaintained and does not work with any recent go-IPFS version.
 
 **Note**: This library occasionally has to change to stay compatible with the IPFS HTTP API.
 Currently, this library is still under testing against [go-ipfs v0.8.0](https://github.com/ipfs/go-ipfs/releases/tag/v0.8.0).
-We strive to support the last 5 releases oftested go-IPFS at any given time; go-IPFS v0.5.0 therefore
+We strive to support the last 5 releases of go-IPFS at any given time; go-IPFS v0.5.0 therefore
 being to oldest supported version at this time.
 
 ## Table of Contents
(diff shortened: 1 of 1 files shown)


c0llab0rat0r:
The code in master is unreleased v0.8.0 support; hence the comment about 0.8.0 in the readme.

The current library in pypi supports v0.7.0.

iskanderandrews:
> The code in master is unreleased v0.8.0 support; hence the comment about 0.8.0 in the readme.
> 
> The current library in pypi supports v0.7.0.

Alright, so I think we are good to close this PR.

PR #278, opened 2021-05-02
Move MatcherSpecInvalidError

### LINKED ISSUE #277: Move MatcherSpecInvalidError into `ipfshttpclient/exceptions.py`
https://github.com/ipfs-shipyard/py-ipfs-http-client/blob/09cae7672a39ca3ff991684c1b31cfe7875174c3/ipfshttpclient/filescanner.py#L416-L420

We currently expose all possible exceptions in the `ipfshttpclient/exceptions.py` module. Ideally this one should be moved there to and added into the existing exception hierarchy (inheriting from both some exception there and `TypeError`, I think).

c0llab0rat0r:
Resolves #277

### DIFF
--- a/ipfshttpclient/exceptions.py
+++ b/ipfshttpclient/exceptions.py
@@ -34,25 +34,27 @@ class AddressError(Error, multiaddr.exceptions.Error):  # type: ignore[no-any-un
 	"""Raised when the provided daemon location Multiaddr does not match any
 	of the supported patterns."""
 	__slots__ = ("addr",)
-	#addr: ty.Union[str, bytes]
+
+	addr: ty.Union[str, bytes]
 	
 	def __init__(self, addr: ty.Union[str, bytes]) -> None:
-		self.addr = addr  # type: ty.Union[str, bytes]
+		self.addr = addr
 		Error.__init__(self, "Unsupported Multiaddr pattern: {0!r}".format(addr))
 
 
 class VersionMismatch(Warning):
 	"""Raised when daemon version is not supported by this client version."""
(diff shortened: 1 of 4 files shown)


ipfshttpclient/exceptions.py
```diff
@@ -97,27 +102,44 @@ class DecodingError(EncoderError):
...
+		self.original = original
 		
 		super().__init__("Object decoding error: {}".format(original), encoder_name)
 
 
+##################
+# filescanner.py #
+##################
+
+class MatcherSpecInvalidError(TypeError):
```
ntninja, line 118: Make this `class MatcherSpecInvalidError(Error, TypeError)` so that will also be caught when run catch-all blocks for any of the library's exceptions (`try: … except ipfshttpclient.exception.Error: …`)

ipfshttpclient/exceptions.py
```diff
@@ -97,27 +102,44 @@ class DecodingError(EncoderError):
...
+	"""
+	An attempt was made to build a matcher using matcher_from_spec, but an invalid
+	specification was provided.
+	"""
+
+	def __init__(self, matcher_class: type, invalid_spec: ty.Any) -> None:
+		super().__init__(
+			f"Don't know how to create a {matcher_class.__name__} from spec {invalid_spec!r}"
+		)
```
ntninja, line 127: This seems a bit nonsensical to me. Won't always `matcher_class.__name__ == "Matcher"`?

ntninja (changes requested):
Thank you, but found a minor thing and a question unfortunately. :slightly_smiling_face:

PR #280, opened 2021-05-11
Prep to release 0.8.0a1

c0llab0rat0r:
* Change version marker from 0.7.0 to 0.8.0a1
* Fix spelling and grammar in change log
* Remove polyfill for `contextlib.suppress` (not needed since Python 3.4)
* Fix 'custom marker not registered' warning related to `pytest.mark.last` (provided by `pytest-ordering` module)

@ntninja I see the project in my PyPi account but it doesn't look like I have any elevated privileges on it.

What are next steps?

### DIFF
--- a/ipfshttpclient/version.py
+++ b/ipfshttpclient/version.py
@@ -8,4 +8,5 @@
 # `0.4.1` and so on. When IPFS `0.5.0` is released, the first client version
 # to support it will also be released as `0.5.0`.
 
-__version__ = "0.7.0"
+__version__ = "0.8.0a1"
+
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -14,19 +14,6 @@ import tempfile
 import pytest
 
 
-if not hasattr(contextlib, "suppress"):
-	"""
-	Polyfill for ``contextlib.suppress``
-	"""
-	@contextlib.contextmanager
-	def _contextlib_suppress(*exceptions):
-		try:
-			yield
-		except exceptions:
-			pass
-	contextlib.suppress = _contextlib_suppress
-
-
 ######################
 # Test configuration #
 ######################
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,18 +3,21 @@ py-ipfs-http-client 0.X.X (XX.XX.20XX)
 
  * (None yet)
 
-py-ipfs-http-client 0.8.0 (XX.XX.20XX)
---------------------------------------
+
+py-ipfs-http-client 0.8.0a1 (11.05.2021)
+----------------------------------------
 
  * py-ipfs-api-client will now only warn when detecting an unsupported daemon version
    * Rationale: During the 0.4.x series the library required many changes to stay compatible
      with the respective daemon version, but since 0.5.0 appears to be no longer the case
- * Compatiblity bumped to go-IPFS 0.8.x (by Jan Rydzewski and other community members bugging me)
-
+ * Compatibility bumped to go-IPFS 0.8.x (by Jan Rydzewski and other community members bugging me)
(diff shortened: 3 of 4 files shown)


test/run-tests.py
```diff
@@ -14,19 +14,6 @@
...
-	"""
-	@contextlib.contextmanager
-	def _contextlib_suppress(*exceptions):
-		try:
-			yield
-		except exceptions:
-			pass
-	contextlib.suppress = _contextlib_suppress
-
-
```
ntninja, line 29: :+1:

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    last: marker supplied by but not registered by pytest-ordering module
```
ntninja, line 3: pytest configuration is already in `tox.ini`. I think your overshadowing that by creating this file.

ipfshttpclient/version.py
```diff
@@ -8,4 +8,5 @@
 # `0.4.1` and so on. When IPFS `0.5.0` is released, the first client version
 # to support it will also be released as `0.5.0`.
 
-__version__ = "0.7.0"
+__version__ = "0.8.0a1"
```
ntninja, line 11: Why 0.8.0 though?

ntninja (commented):
Thanks for the spelling fixes! Is the pre-release on purpose? I'm not planning on adding anything for this cycle anymore…

ntninja:
> @ntninja I see the project in my PyPi account but it doesn't look like I have any elevated privileges on it.

You need to publish it using `flit publish` or `twine`, you cannot publish it directly through the website. ([Official publishing docs](https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives), but use `flit publish` where it says `twine`.)

pytest.ini
c0llab0rat0r: I was getting a warning running the build; adding this file fixed the warning.

ipfshttpclient/version.py
c0llab0rat0r: I saw you had done that with a `0.7` previously so I followed suit. Also erring on the side of caution, since this would be my first time running the publish process.

Shouldn't be hard to republish `0.8.0` after publishing `0.8.0a1` if all is well with `0.8.0a1`.

pytest.ini
ntninja: I'm just saying you should just add the “markers” setting to the respective section in *tox.ini* instead to avoid the extra file and because I think pytest will ignore the other configuration already in *tox.ini* if you create this one.[citation needed]

ipfshttpclient/version.py
ntninja: Fair enough :slightly_smiling_face:

dokterbob:
What's blocking this gentle people? :)

c0llab0rat0r:
> What's blocking this gentle people? :)

We have a bottleneck / dependency in that @ntninja is the only one with full access to the repository and release process.

ntninja:
Looking good now. Thanks!

ntninja:
@c0llab0rat0r: Are you able to push this to PyPI now?

c0llab0rat0r:
> @c0llab0rat0r: Are you able to push this to PyPI now?

Released as https://pypi.org/project/ipfshttpclient/0.8.0a1/ (project home page only shows regular releases so it shows 0.7.0).

Also pushed a signed tag to my fork:

https://github.com/c0llab0rat0r/py-ipfs-http-client/releases/tag/0.8.0a1

c0llab0rat0r:
I don't have the IPNS private key, so I can't update the documentation.

PR #281, opened 2021-05-18
v0.8.0a2

c0llab0rat0r:
* Documentation corrections
* Support publishing documentation to a remote IPFS server
* Remove pytest skips related to old versions of Python and old versions of IPFS
* Move release documentation to RELEASE.md in git root (first place people would look for it; also fixes a Sphinx warning)
* Update release documentation
* Fix remaining Sphinx warnings
* Configure Sphinx to fail on warnings


c0llab0rat0r:
@ntninja

docs/publish.py
```diff
@@ -1,41 +1,76 @@
 #!/usr/bin/python3
 import os
 import sys
-__dir__ = os.path.dirname(__file__)
-sys.path.insert(0, os.path.join(__dir__, ".."))
+import typing as ty
+
+script_dir = os.path.dirname(os.path.realpath(__file__))
+sys.path.insert(0, os.path.join(script_dir, ".."))
```
ntninja, line 7: IMHO its almost always nicer to use `pathlib.Path`, rather then `os.path` when dealing with filesystem paths:

```py
__dirpath__ = pathlib.Path(__file__).resolve().parent
sys.path.insert(0, __dirpath__ / "..")
```

(Just a suggestion though.)

docs/publish.py
```diff
@@ -1,41 +1,76 @@
...
-		print("Usage: {0} [IPNS-key]".format(os.path.basename(program)))
+		print("Usage: {0} [IPNS-key]".format(os.path.basename(__file__)))
 		print()
 		print("!! Continuing without publishing to IPNS !!")
 		print()
-	
+
+	ipfs_api_address: str = os.getenv('IPFS_API_MULTI_ADDR', str(ipfshttpclient.DEFAULT_ADDR))
+	ipfs_api_username: ty.Optional[str] = os.getenv('IPFS_API_USERNAME', None)
+	ipfs_api_password: ty.Optional[str] = os.getenv('IPFS_API_PASSWORD', None)
```
ntninja, line 28: Using environment variables seems weird to me here. This would make a lot more sense as command-line parameters instead?

docs/publish.py
```diff
@@ -1,41 +1,76 @@
...
-	result = sphinx.cmd.build.build_main(["-b", "html", "-d", "build/doctrees", ".", "build/html"])
+	result = sphinx.cmd.build.build_main([
+		"-b",
+		"html",
+		"-d",
+		"build/doctrees",
+		".",
+		"build/html",
+		"-W",
+		"--keep-going"
```
ntninja, line 52: I suggest not putting everything on seperate lines for readability:

```py
"-b", "html",
"-d", "build/doctrees",
".", "build/html",
"-W", "--keep-going",
```

docs/publish.py
c0llab0rat0r: Using environment variables for credentials is very common (see Docker).

docs/publish.py
c0llab0rat0r: `pathlib.Path` appeals to me also, I was maintaining the status quo here.

docs/publish.py
```diff
@@ -1,41 +1,76 @@
...
+def main(argv: ty.List[str]) -> int:
+	if len(argv) >= 1:
+		ipns_key = argv[0]
+	else:
+		ipns_key = None
 
-def main(argv=sys.argv[1:], program=sys.argv[0]):
-	if len(argv) != 1:
-		print("Usage: {0} [IPNS-key]".format(os.path.basename(program)))
+		print("Usage: {0} [IPNS-key]".format(os.path.basename(__file__)))
```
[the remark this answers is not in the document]
c0llab0rat0r: There was a bug where the script could not be executed directly from its own folder and had to be referenced as `docs/publish.py` instead of `publish.py` - that is fixed in this PR.

ipfshttpclient/http.py
```diff
@@ -30,4 +33,7 @@
 	except ImportError:  # pragma: no cover
 		from . import http_httpx as _backend
 
-ClientSync = _backend.ClientSync
+
+# noinspection PyPep8Naming
+def ClientSync(*args, **kwargs) -> ClientSyncBase[ty.Any]:  # type: ignore[no-untyped-def]
+	return _backend.ClientSync(*args, **kwargs)
```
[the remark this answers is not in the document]
c0llab0rat0r: Given that `ClientSync` is a private implementation detail of the library, and only called from `class ClientBase` - we arn't constrained to support backward compatibility on it like we would a class such as `Client`.

The gain here is a warning-free `Sphinx` documentation generation, which enables warnings to be configured as errors, thus catching problems during the build and publish that would otherwise be caught after docs have been made public.

As far as the specific approach I took (`def ClientSync(*args ..etc)`) I agree it is.... not pretty. I explored other approaches before choosing this one; they either broke Sphinx, or they required a larger code change to the rest of the library.

The current approach minimizes the amount of code change. The larger changes involved renaming `ClientSync` to look like a factory/builder method (small/moderate change) or to introduce another abstract base class (large change, seemed like a solution worse than the original problem).

Do we need to maintain two http client implementations going forward? Is that an artifact of supporting Python versions that are no longer supported (such as Python 3.5)?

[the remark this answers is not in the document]
c0llab0rat0r: > This is a bad idea, since people can do:
> 
> ```python
> class MyClient(ipfshttpclient.Client):
>     def some_extra_method(self):
>         …
> ```
> 
> and this change would break that IMHO for no gain at all.

`ipfshttpclient.Client` does not inherit from `ClientSync`:

`class Client(files.Base, miscellaneous.Base):`

`files.Base` inherits from `class Base(base.ClientBase):` which creates an instance of `ClientSync` (or in this PR, calls it as a factory method). I.e., `ClientBase` is getting a `ClientSync` via composition instead of inheritance.

c0llab0rat0r:
@ntninja I made some changes based on your feedback and added replies to others.

docs/publish.py
ntninja: It feels wrong to do it this way, but I get your point. Could you drop `IPFS_API_MULTI_ADDR` in favour of `PY_IPFS_HTTP_CLIENT_DEFAULT_ADDR` and add a new `PY_IPFS_HTTP_CLIENT_DEFAULT_AUTH` instead of the `IPFS_API_USERNAME` and `IPFS_API_PASSWORD` fields?

At least that way passing the envs this way not be special to just this specific piece of maintenance code.

ipfshttpclient/http.py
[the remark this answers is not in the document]
ntninja: > The gain here is a warning-free Sphinx documentation generation, which enables warnings to be configured as errors, thus catching problems during the build and publish that would otherwise be caught after docs have been made public.

What kind of warning is it anyways? Assuming its a dumb one that doesn't help anybody: Can't we just ignore the warning instead?

> Do we need to maintain two http client implementations going forward? Is that an artifact of supporting Python versions that are no longer supported (such as Python 3.5)?

Well, HTTPx never supported Python 3.5 so that's why we have this solution in the first place. I'm a bit hasitant to go all-in on HTTPx through until it gets out of beta – its APIs are still theoretically subject to change after all.
This change was actually done in preparation of adding `async`-support to the library – something that requests doesn't support and may never will.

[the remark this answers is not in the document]
c0llab0rat0r: Thanks for the explanation on HTTPx, I understand that better now.

c0llab0rat0r:
@ntninja I've applied changes based on your second round of feedback.

ntninja (approved):
Thank you!

c0llab0rat0r:
@ntninja 

I've published `0.8.0a2` to PyPi.

It's tagged here: https://github.com/c0llab0rat0r/py-ipfs-http-client/releases/tag/0.8.0a2
And the previous: https://github.com/c0llab0rat0r/py-ipfs-http-client/releases/tag/0.8.0a1

I don't have merge rights on the repo so I can't tag the releases.

ntninja:
Pushed the tags! I've reached out to have you added to the GitHub team eventually, but I cannot do it myself unfortunately.

Luflosi:
I think you accidentally pushed the tags to https://github.com/ntninja/py-ipfs-api-client/releases, not to this repo. Can you please also push them to this repo?

ntninja:
@Luflosi: Done, sorry 'bout that again! :-/


Am Sonntag, 27. Juni 2021, 18:23:48 CEST schrieb 
Luflosi:
> I think you accidentally pushed the tags to
> https://github.com/ntninja/py-ipfs-api-client/
releases, not to this repo.
> Can you please also push them to this repo?

Luflosi:
Thank you.

PR #286, opened 2021-06-24
Replace pytest_ordering with pytest_order

Luflosi:
pytest_ordering seems unmaintained as it hasn't been updated in two years: https://github.com/ftobia/pytest-ordering. It is also broken, at least in Nixpkgs: https://github.com/NixOS/nixpkgs/pull/122264.
pytest_order is a maintained and working fork.
I added a similar patch to Nixpkgs a while ago to make it work there: https://github.com/NixOS/nixpkgs/blob/6965eef223f491cc1d9289ce972d8ab6811a2d28/pkgs/development/python-modules/ipfshttpclient/default.nix#L59-L63. It would be nice to get rid of this patch for a future version.
I bumped the version in tox.ini a little since version 0.6 does not appear in https://pypi.org/project/pytest-order/#history.

I'm not sure if the line with `markers =` in tox.ini is still correct with this change, please check.

### DIFF
--- a/test/functional/test_miscellaneous.py
+++ b/test/functional/test_miscellaneous.py
@@ -21,7 +21,7 @@ def test_id(client):
 # Shutdown test #
 #################
 
-@pytest.mark.last
+@pytest.mark.order("last")
 def test_daemon_stop(daemon, client):
 	# The value for the `daemon` “fixture” is injected using a pytest plugin
 	# with access to the created daemon subprocess object defined directly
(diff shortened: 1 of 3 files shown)


ntninja:
Thanks! The marker was indeed only used to work around `pytest-ordering`s failure to register its own markers. Simply hadn't checked for alternatives to this plugin in a long while as it continued working and this workaround was very minimal.

c0llab0rat0r:
Thanks for this library cleanup.

PR #285, opened 2021-06-14
[wip] Fix bug with httpx usage - ensure basic auth is set when provided

c0llab0rat0r:

PR #289, opened 2021-06-30
Fix authentication bug with httpx usage bug

c0llab0rat0r:
And:

* add `status_code` property to `StatusError` class
* change `PartialErrorResponse` to inherit from `Error` instead of `ErrorResponse` due to lack of success status code availablilty at call site throwing `PartialErrorResponse`

This PR is an alternative to #285.


c0llab0rat0r:
@ntninja this is ready for review.

PR #290, opened 2021-06-30
Fix typing problems with http_requests and requests_wrapper

c0llab0rat0r:
Most remaining issues (after this pr) are due to:

* urllib3 not having enough type info
* mypy not supporting mixins well (or perhaps we are not using mixins correctly with mypy)


c0llab0rat0r:
@ntninja this is ready for review but I don't think it should be merged until after the `0.8.0` release since its somewhat of an invasive refactor.

#289 should go first be merged for 0.8.0.

PR #294, opened 2021-08-19
Update go-ipfs version to support < 0.10

kclowes:


kclowes:
Tests are passing locally for me, and I'm not sure how to go about debugging on Travis. Let me know if anyone has any ideas as to why it might be failing!

jacekv:
I checked with IPFS version 0.9.1 and I get the same problem with the OSX tests. It seems there is a Python issue: https://bugs.python.org/issue32958
I am currently having a look if there is something we can do to fix this issue.

jacekv:
I can't find a way to test this besides running the pipeline all the time. I don't have a Mac. So maybe someone else will be able to have a look into this issue :/

hexcowboy:
The tests are passing on my MacBook.

```bash
platform darwin -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-1.0.0 -- /py-ipfs-http-client/.tox/py3/bin/python
---
test/unit/test_http.py::test_successful_request_uds PASSED                                                                [ 34%]
test/unit/test_http.py::test_generic_failure_uds PASSED                                                                   [ 35%]
```

```bash
$ python --version
Python 3.9.7
```

There is no reason they should not pass. Apple macOS is Unix compliant and supports Unix Domain Sockets (UDS). Something is either wrong with CI or Python 3.9.0, most likely the former.

jacekv:
Interesting. I tried it again but the error remains: https://app.travis-ci.com/github/ipfs-shipyard/py-ipfs-http-client/jobs/538855219
Using Python 3.9.7 and the same error is there. Maybe it is something with the CI?

hexcowboy:
@jacekv I'm not familiar with Travis CI. Can you try with a version of macOS 11.x? The current release is 11.6. If this continues to fail I would recommend the project maintainers remove the UDS test for macOS specific builds.

jacekv:
@hexcowboy  Good point regarding the os version. I changed it to 11.3, since this is what seems to be currently supported by Travis and the error is still there.
Might make sense to remove the UDS tests for OSX.

kallisti5:
Is this ever going to get worked on?   IPFS is now on 0.13.0 and this only supports up to version 0.8.0 :-|

PR #323, opened 2023-05-17
removed the pubsub functionality

Wizock:
Changes:
- Removed pubsub.py
- Removed imports from other files


kasteph (changes requested):
Thanks for contributing @Wizock!

It looks like your editor also formatted files that are irrelevant for this particular issue (thus making the PR larger and harder to read which parts should actually be reviewed). Could you disable that first? Thanks!

